PR #117, opened 2015-12-09
Close  #115: Allow user to store non-openPMD information

### LINKED ISSUE #115: Allow user to store non-openPMD information
Today I had a request from a prospective user of openPMD: can the user store other datasets in the HDF5-openPMD files, which are **not meant** to be read by the openPMD parser / viewer ? (for instance because they don't fall either in the category mesh or or in the category particle)

From my point of view, the best way to do this is to store this data outside of the `basePath` (in this case the parser will not try to read it). 

@ax3l : Does that make sense ? If yes, I think it would be good to add it as a side-remark in the standard.

RemiLehe:
Close #115: This allows user to store additional data, outside of the basePath.

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -99,7 +99,18 @@ Each file's *root* directory (path `/`) must at leat contain the attributes:
                    to create a real path from it replace all occurrences
                    of `%T` with the integer value of the iteration, e.g.,
                    `/data/%T` becomes `/data/100`
-    - allowed value: fixed to `/data/%T/` for this version of the standard
+    - allowed value: fixed to `/data/%T/` for this version of the
+    standard
+	- remark: all the data that is formatted according to the present
+      standard (i.e. both the meshes and the particles) is to be
+      stored within a path of the form given by `basePath` (e.g. in
+      the above example, the data will be stored within the path `/data/100/`).
+	  If, for various reasons, a user wants to store **additional
+      data** that is not (or cannot be) formatted according to the
+      present standard (e.g. fields on an unstructured mesh),
+      this can be done be storing this data within a path that **is not**
+      of the form given by `basePath` (e.g. `/extra_data`). In this
+      way, the openPMD parsing tools will not parse this additional data. 
 
   - `meshesPath`
     - type: *(string)*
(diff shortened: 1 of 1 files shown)


ax3l:
thx for the PR! :)

STANDARD.md
```diff
@@ -100,6 +100,16 @@ Each file's *root* directory (path `/`) must at leat contain the attributes:
                    of `%T` with the integer value of the iteration, e.g.,
                    `/data/%T` becomes `/data/100`
     - allowed value: fixed to `/data/%T/` for this version of the standard
+	- remark: all the data that is formatted according to the present
```
ax3l: pls no tab :)
can you use "note: " instead? :)

STANDARD.md
```diff
@@ -100,6 +100,16 @@ Each file's *root* directory (path `/`) must at leat contain the attributes:
...
+      standard (i.e. both the meshes and the particles) is to be
+      stored within a path of the form given by `basePath` (e.g. in
+      the above example, the data will be stored within the path `/data/100/`).
+      If, for various reasons, a user wants to store **additional
+      data** that is not (or cannot be) formatted according to the
+      present standard (e.g. fields on an unstructured mesh),
+      this can be done be storing this data within a path that **is not**
+      of the form given by `basePath` (e.g. `/extra_data`). In this
```
ax3l, line 111: we could write more simple: "within a path that does **not reside** inside `basePath` (e.g. `/extra_data`)."

STANDARD.md
```diff
@@ -100,6 +100,16 @@ Each file's *root* directory (path `/`) must at leat contain the attributes:
                    of `%T` with the integer value of the iteration, e.g.,
                    `/data/%T` becomes `/data/100`
     - allowed value: fixed to `/data/%T/` for this version of the standard
+	- remark: all the data that is formatted according to the present
+      standard (i.e. both the meshes and the particles) is to be
+      stored within a path of the form given by `basePath` (e.g. in
+      the above example, the data will be stored within the path `/data/100/`).
+      If, for various reasons, a user wants to store **additional
```
ax3l: formatting-wise: I think we currently _stress_ important facts that do not introduce keywords, etc. only with a simple italic `*stressed fact*` in the text.

RemiLehe:
Thanks for the comments! I updated the PR.

PR #118, opened 2016-01-06
fix python code example

skuschel:
checked with openPMD example data

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -438,22 +438,22 @@ short-hand notation (see: *Constant Record Components*).
                               with `constant record components`
     - example: reading example (with h5py) in Python:
 ```python
-def is_const_record(record_name, component_name) :
-    return ("value" in record_name[component_name].attrs.keys())
+def is_const_record(record, component_name) :
+    return ("value" in record[component_name].attrs.keys())
 
-def get_component(record_name, component_name) :
-    if is_const_record(record_name, component_name) :
-        return record_name[component_name].attrs["value"]
+def get_component(record, component_name) :
+    if is_const_record(record, component_name) :
+        return record[component_name].attrs["value"]
     else :
-        record_name[component_name].value
+        return record[component_name].value
 
 f = h5py.File('example.h5')
 species = f["<path_to_species_group>"]
 
 position_x_relative = get_component(species["position"], "x") \
-                    * species["position"].attrs["unitSI"]
+                    * species["position/x"].attrs["unitSI"]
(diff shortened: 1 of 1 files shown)


ax3l:
@skuschel thank you for the PR!

can you pls direct anything new against the `upcoming-X` branch(es), in this case `upcoming-1.0.1`?
`latest` is similar to a [`master` branch](http://nvie.com/posts/a-successful-git-branching-model/) that does not get direct updates unless it is a release (which is immediately tagged, documented, archived for a DOI, etc.)

I will now have a look inside the PR, stay tuned.

STANDARD.md
```diff
@@ -438,22 +438,22 @@ short-hand notation (see: *Constant Record Components*).
...
     else :
-        record_name["component_name"][:]
+        return record[component_name].value
 
 f = h5py.File('example.h5')
 species = f["<path_to_species_group>"]
 
 position_x_relative = get_component(species["position"], "x") \
-                    * species["position"].attrs["unitSI"]
+                    * species["position/x"].attrs["unitSI"]
```
ax3l, line 454: yes, that was outdated, good catch!

STANDARD.md
```diff
@@ -438,22 +438,22 @@ short-hand notation (see: *Constant Record Components*).
                               with `constant record components`
     - example: reading example (with h5py) in Python:
 ```python
-def is_const_record(record_name, component_name) :
-    return ("value" in record_name["component_name"].attrs.keys())
+def is_const_record(record, component_name) :
```
ax3l, line 441: not sure if that change is necessary, but I would be fine with it since it is actually a `record` _object_ and not a _name_, true.

ax3l:
content looks great! please rebase against our _mainline_ `upcoming-1.0.1` branch and direct your PR against it! :)

thanks, well spotted! :sparkles:

STANDARD.md
skuschel, line 441: this was just readability: both are suffixed `_name`, but the record was actually the hdf5 object, but `component_name` is a string.

skuschel:
sure, changing the branch makes sense! I didnt think about it at all, just used the `default` branch, github was suggesting (see: settings -> Branches)...

ax3l:
yes, `latest` is our default since we want new readers to end up there (that might not be git experts).
That's why we as devs (with more git knowledge) have to select the right one :)

skuschel:
too late for reading that last night :)
however, I sent a new PR

PR #119, opened 2016-01-07
Particle Position Reading: Fix Python Example

skuschel:
Fix the outdated record reading example (python).

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -448,22 +448,22 @@ short-hand notation (see: *Constant Record Components*).
                               with `constant record components`
     - example: reading example (with h5py) in Python:
 ```python
-def is_const_record(record_name, component_name) :
-    return ("value" in record_name[component_name].attrs.keys())
+def is_const_record(record, component_name) :
+    return ("value" in record[component_name].attrs.keys())
 
-def get_component(record_name, component_name) :
-    if is_const_record(record_name, component_name) :
-        return record_name[component_name].attrs["value"]
+def get_component(record, component_name) :
+    if is_const_record(record, component_name) :
+        return record[component_name].attrs["value"]
     else :
-        record_name[component_name].value
+        return record[component_name].value
 
 f = h5py.File('example.h5')
 species = f["<path_to_species_group>"]
 
 position_x_relative = get_component(species["position"], "x") \
-                    * species["position"].attrs["unitSI"]
+                    * species["position/x"].attrs["unitSI"]
 position_x_offset = get_component(species["positionOffset"], "x") \
-                  * species["positionOffset"].attrs["unitSI"]
+                  * species["positionOffset/x"].attrs["unitSI"]
 
 x = position_x_relative + position_x_offset
 ```
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -449,13 +449,13 @@ short-hand notation (see: *Constant Record Components*).
     - example: reading example (with h5py) in Python:
 ```python
 def is_const_record(record_name, component_name) :
-    return ("value" in record_name["component_name"].attrs.keys())
+    return ("value" in record_name[component_name].attrs.keys())
 
 def get_component(record_name, component_name) :
     if is_const_record(record_name, component_name) :
-        return record_name["component_name"].attrs["value"]
+        return record_name[component_name].attrs["value"]
     else :
-        record_name["component_name"][:]
+        record_name[component_name].value
 
 f = h5py.File('example.h5')
 species = f["<path_to_species_group>"]


STANDARD.md
```diff
@@ -448,22 +448,22 @@ short-hand notation (see: *Constant Record Components*).
...
 
-def get_component(record_name, component_name) :
-    if is_const_record(record_name, component_name) :
-        return record_name["component_name"].attrs["value"]
+def get_component(record, component_name) :
+    if is_const_record(record, component_name) :
+        return record[component_name].attrs["value"]
     else :
-        record_name["component_name"][:]
+        return record[component_name].value
```
ax3l, line 458: ah, interesting. I never used `.value` on `h5py.Dateset` types before. It returns the `numpy.ndarray` directly, cool. (`Docstring:       Alias for dataset[()]`)
yes, that is way more general then the example assuming a 1D array `[:]`. Alternatively, `[()]` would be the same says the docstring.

STANDARD.md
```diff
@@ -448,22 +448,22 @@ short-hand notation (see: *Constant Record Components*).
...
 
 f = h5py.File('example.h5')
 species = f["<path_to_species_group>"]
 
 position_x_relative = get_component(species["position"], "x") \
-                    * species["position"].attrs["unitSI"]
+                    * species["position/x"].attrs["unitSI"]
```
ax3l, line 464: good catch: `unitSI` is on the components, `unitDimension` on the record (which is the same object only for scalar records).

ax3l:
@skuschel looks good to me, I let @RemiLehe have a look too and then we can merge it.

Thanks a lot for fixing the example, that is great! :sparkles:

STANDARD.md
ax3l, line 464: access should work for `"position/x"` being either a hdf5 group or a data set (good, no special treatment for constant records necessary at this point).

skuschel, line 464: right, there is no special treament necessary. I generalized it for the postpic reader like this: https://github.com/skuschel/postpic/blob/openPMDreader/postpic/datareader/openPMDh5.py#L73

However, i didnt get my head aroud, why you guys build it that way: a scalar value is stored in `dataset.attrs['value']`, but an array in `dataset.value`. Why doesnt `dataset.value` return a float or an array with only a single element in that case?

ax3l, line 464: nope, it's not `scalar` in that sense. when we say `scalar` (1) or `vector` (>1) we mean the number of [**components** of an **openPMD record**](https://github.com/openPMD/openPMD-standard/blob/upcoming-1.0.1/STANDARD.md#scalar-vector-and-tensor-records).

[Examples:](http://www.openpmd.org/#/step-16) temperature is a scalar record, momentum is a vector record.

what you see in `record["constant_component_name"].attrs['value']` is a **constant** record component, which is an orthogonal concept that allows to save memory. It replaces the `h5py.Dataset` that would be stored at this position with a `h5py.Group` (+attributes such as `value`, `shape`).

Just returning a single value would be very problematic if one wants to determine the actual (hidden) total number of elements (aka the `shape`) and while doing _element-wise operations_ on _several components of a record_ that are expected to have the _same lengths_ (e.g., only `position/z` is constant).

As a reader application, if one wants to remove that division on the cost of memory, one can at read of a constant component allocate a new `numpy.ndarray` of size `shape` with all elements being initialized to the `value` attribute. Nevertheless, there are applications (also in other programming languages) that might want to keep that level of abstraction for the sake of performance and because they treat constant ([r-values](https://en.wikipedia.org/wiki/Value_%28computer_science%29#Assignment:_l-values_and_r-values)) vs. varying (l-values) record components also completely different.

To make it even more complicated: constant record components are designed to be _fully meta data_ in HDF5/ADIOS which is even more ideal since they are cached and read/written at once (that's why we use groups and attributes for them, not single-entry data sets(h5)/variables(bp)).

The second question on `dataset.value` is just `h5py` syntax to access the values inside a `h5py.Dataset` object. Luckily it is the same as our naming choice for the attribute `value` that is added for constant record components, making things nice :)

ax3l, line 464: I think your linked [data](https://github.com/skuschel/postpic/blob/a6f8590f6ee65cb2c604745f18b2d2640ce0b54d/postpic/datareader/openPMDh5.py#L73-L81) method is mixing the [naming](https://github.com/openPMD/openPMD-standard/blob/upcoming-1.0.1/STANDARD.md#scalar-vector-and-tensor-records) of **openPMD record** and **openPMD component** (e.g., in the variable names). I think `record` should be named `record_component` for instance. Can you add a docstring what `data` actually returns (_values of record components_)?

Why do you cast the return type to `np.float64`?

skuschel, line 464: the data method is only meant to be used on **components** (`B/x` or 'temperature', but not `B`).
`electrons/positions/x` usually returns an array of length `N` (number of particles). postpic treats every particle property like that. For `electrons/mass` this is usually a wrong assumption, but what you will get instead is an array with only a single float in it. The optimizations you are talking about (element-wise against constant) are now automatically performed in numpy and postpic doesnt have to worry about it.

Since postpic can do a lot of calculations with your data (if you want it to), not converting to float64 ended in rounding errors visible in the final result in some rare cases. I just dont want to bother and its actually a fast operation compared to reading from HDD, histogram creation or saving the fingure with matplotlib.

ax3l, line 464: sounds great, all right!

skuschel, line 464: thanks for your feedback!

RemiLehe:
The corrections are fine with me. @skuschel : Thanks a lot for spotting and correcting these mistakes !
@ax3l : You can merge it.

PR #127, opened 2016-05-10
Refactor Python position Example

ax3l:
Second change for the example how to combine `position` and `positionOffset` to make it more general.

The functions provided now are corrected in namings and usable for general `record components` even outside of the `particle` scope (e.g., one can use them to read scalar and vector `mesh` components).

I still did not add doc strings since I try not to blow up the _inline_ example since it is already pretty large. 
## 

_outdated_: The `allocate` option is quite of a fancy thing in case one wants to transparently allocate an ndarray to access it via `[]` later on but could be removed.

Otherwise I can also keep it like this:

``` python
def get_component(group, component_name):
    record_component = group[component_name]
    unitSI = record_component.attrs["unitSI"]

    if is_const_record_component(record_component):
        return record_component.attrs["value"], unitSI
    else:
        return record_component.value, unitSI
```

instead of

``` python
def get_component(group, component_name, allocate=False):
    record_component = group[component_name]
    unitSI = record_component.attrs["unitSI"]

    if is_const_component(record_component):
        cvalue = record_component.attrs["value"]
        if allocate:
            return cvalue * \
                numpy.ones(shape=record_component.attrs["shape"],
                           dtype=record_component.attrs["value"].dtype),
                unitSI
        else:
            return cvalue, unitSI
    else:
        return record_component.value, unitSI
```

update: yes, that is probably better... too fancy and not necessary to know for most people I guess...

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -428,6 +428,18 @@ short-hand notation (see: *Constant Record Components*).
 
 ### Records for each `Particle Species`
 
+  - `id`
+    - type: *(uint64 / UNSIGNED8)*
+    - scope: *optional*
+    - description: a globally-unique identifying integer for each particle,
+                   that can be used to, e.g., track particles. This
+                   identifying integer should be truly unique within the
+                   simulation; in particular, even among different particle
+                   species, two particles should not have the same id.
+                   Also, when a particle exits the simulation box, its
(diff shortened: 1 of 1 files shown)


RemiLehe:
Great. Thanks for the update. 
I also prefer it without the allocate option.

PR #131, opened 2016-06-14
Typo: Allowed Values for `fieldSmoothing`

ax3l:
Fixes the reference for allowed values in `fieldSmoothing` (same as in current).

Discovered copy-paste typo by @Flamefire, thanks!


RemiLehe:
Thanks for spotting this!

PR #133, opened 2016-07-06
ED-PIC: Fix PSATD Long Name

ax3l:
Fix a copy & paste error in the pseudo-spectral _analytical_ time-domain (PSATD) algorithm description.


RemiLehe:
Good catch!

Issue #146, opened 2017-11-20
Rename (float) to (floatX)


ax3l:
The conventions in the first section of the paragraph are confusing since many people think `(float)` is always single precision.

Change it to append `X`.

Maybe also rename `double` with `float64`.

PR #152, opened 2017-11-24
Unspecified float/int append X

ax3l:
This is a wording change of the openPMD 1.0.0 standard as proposed in #146.

## Description

When we are referring to floating point numbers with unspecified (user-chosen) precision, we use `(floatX)` for `(float16|32|64|128|...)` now to avoid confusion with single precision "C" floats. The same change is added for the type of (unsigned) integers.

Also replaces the usage of C/C++/Python `double` with `float64` for consistency.

## Affected Components

- `base`
- extension: `ED-PIC`

## Logic Changes

No changes introduced.

## Writer Changes

No data writers affected.

## Reader Changes

No data readers affected.

## Data Converter

No data conversion needed.

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -12,10 +12,12 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 
 All `keywords` in this standard are case-sensitive.
 
-The naming *(floatX)* without a closer specification is used if the implementor
-can choose which kind of floating point precision shall be used.
-The naming *(uintX)* and *(intX)* without a closer specification is used if the
-implementor can choose which kind of (un)signed integer type shall be used.
+The naming *(floatX)* without further specification is used if the implementor
+can choose which kind of floating point precision shall be used
+(e.g. *(float16)*, *(float32)*, *(float64)*, *(float128)*, etc.).
+The naming *(uintX)* and *(intX)* without further specification is used if the
+implementor can choose which kind of (un)signed integer type shall be used
+(e.g. *(int32)*, *(uint64)*, etc.).
 The naming for the type *(string)* refers to fixed-length, plain ASCII encoded
 character arrays since they are the only ones that are likely to propagate
 through all file-format APIs and third-party programs that use them.
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -12,9 +12,9 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 
 All `keywords` in this standard are case-sensitive.
 
-The naming *(float)* without a closer specification is used if the implementor
+The naming *(floatX)* without a closer specification is used if the implementor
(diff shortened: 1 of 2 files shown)


ax3l:
@RemiLehe ping :) please review & merge me :)

STANDARD.md
```diff
@@ -12,9 +12,9 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 
 All `keywords` in this standard are case-sensitive.
 
-The naming *(float)* without a closer specification is used if the implementor
+The naming *(floatX)* without a closer specification is used if the implementor
```
RemiLehe: I think the word *closer* is ambiguous here. I guess you mean `without any further specification`. But this applied to `float` (which did not have any further specification, like `float16`). With the new notation `floatX`, I am not sure the phrase `without a closer specification` makes sense anymore, it the sense that `X` is, in a way, a specification.

STANDARD.md
```diff
@@ -12,9 +12,9 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
...
 can choose which kind of floating point precision shall be used.
```
RemiLehe: Here, just for clarification, I would add examples:
```
can choose which kind of floating point precision shall be used (e.g. float32, float64, float128, etc.)
```

STANDARD.md
```diff
@@ -12,9 +12,9 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
...
-The naming *(uint)* and *(int)* without a closer specification is used if the
+The naming *(uintX)* and *(intX)* without a closer specification is used if the
```
RemiLehe: The above comments for `float` also apply for `int` and `uint`

RemiLehe (changes requested):

RemiLehe:
Thanks for the clarifications in this pull request. I have a few minor requests (see inline comments)

STANDARD.md
ax3l: That's a great point, we could maybe even shorten it to just `without further specification`?

STANDARD.md
ax3l: Good point, makes it less abstract.

ax3l: It would probably make sense to then rename all the `double` namings to `float64` explicitly as we already do it for the Fortran equivalents.
C++ is not to sane in defining it's "double" definition and we really do mean IEEE-754 float64 anyway where used.

ax3l:
Thank you for the comments, I adjusted the wording accordingly to remove ambiguity and added the examples.

For consistency, I also renamed the few usages of `double` for IEE-754 double precision with `float64` (only used for conversion to absolute units in SI).

RemiLehe (approved):

RemiLehe:
Thanks for updating the PR! I'll merge now.

Issue #167, opened 2017-12-20
meshName only mentioned once in the standard 


DavidSagan:
In the 1.0.0 standard:

> The following attributes must be stored additionally with the `meshName` record...

But `meshName` is nowhere else mentioned in the standard. This is very confusing and should be corrected.

PR #168, opened 2018-01-23
Fix meshName: Means mesh record

ax3l:
Do not introduce the name `meshName` if it is only used once and actually means "mesh record".

*Implements issue:* #167

## Description

In the base standard, the naming `meshName` was unnecessarily introduced for a simple mesh record without using it later on.

## Affected Components

- `base`

## Logic Changes

None.

## Writer Changes

No effects.

## Reader Changes

No effects.

## Data Converter

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -295,7 +295,7 @@ homogenous records, usually in a N-dimensional matrix.
 
 ### Required Attributes for each `mesh record`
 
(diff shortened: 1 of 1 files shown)


RemiLehe (approved):

ax3l:
@DavidSagan since you reported #167, does this PR address it ok for you? :)

DavidSagan (approved):

Issue #160, opened 2017-12-10
Use "group" instead of "directory"


DavidSagan:
In the standard "group" and "directory" are used interchangeably. It would make things clearer if only one or the other was used and "group" makes more sense than "directory".

@ax3l: Please put this in the project 1.1 queue.

ax3l:
Thanks for the report!

Actually, only used to refer to `/`. The term "directory" is currently used three times in exactly that formulation:

> *root* directory (path `/`)

when referring to `/`.

I agree that we can also call this immediately 

> *root* group (path `/`)

.

ax3l:
implemented

PR #169, opened 2018-01-23
Root Path: Call Group

ax3l:
Call the root `/` directory group like all other in-file groups.

*Implements issue:* #160

## Description

At a few initial occasions, the root "path" `/` of a openPMD file was referred to as "directory". We now call it consequently a "group" as in all other places.

Also fixes a typo in "lea**s**t".

## Affected Components

- `base`

## Logic Changes

None.

## Writer Changes

No effects.

## Reader Changes

No effects.

## Data Converter


RemiLehe (approved):

RemiLehe:
Thanks for updating this!

DavidSagan (approved):

Issue #143, opened 2017-03-07
How to *not* have a `particlesPath`?


mccoys:
I am wondering how I could manage to *not* have a `particlesPath` in a file. It seems that the `checkOpenPMD_h5.py` tool is always looking for one, and searching it for all required attributes.

Is there a way to prevent that?

And what about having no `meshesPath` instead?

ax3l:
Thank you for the question!

For now, just define it to an empty (or non-existing) folder.
Example: `meshesPath = "fields/"`, `particlesPath = "noParticles/"`

Readers looking for particles will check that path and see that the iteration contains indeed no particle species.

ax3l:
ping @RemiLehe: what you asked today

I would suggest something like this in 1.0.1: Add a note at this point that each of the pathes might not exist and this needs to be handled as "no meshed (particles)" found.

We can decide if readers shall or must not issue a warning (but maybe a less important debug level such as info) in such a case. We could also say one should throw a warning but continue, if the path does not exist but throw not a warning if the path exists but no entries are found. But I guess no warning is the most reasonable way to handle this.

ax3l:
discussion today:
even better: make the attributes optional instead of writing rubbish for 2.0 (but then the paths are required to exist if the attribute exists)

ax3l:
@RemiLehe @C0nsultant I would suggest I fix the remaining `1.0.1` wording changes and make this an immediate follow-up `1.1.0`.

I would immediately relax both paths to be `optional` instead of allowing writing "something". This allows to implement it in yt, checker the viewer, etc. simply and together with the `1.0.X` implementations and makes it more clear what is happening.

I can also provide PRs for all three with the PR.

anokfireball:
Admittedly, I still do not see how the proposed combination of required attribute and optional concrete path is the more sensible option for `1.0.X` (i.e. how this better represents the file in openPMD metadata than enforcing the path). If that is what it takes to not break a majority of the already existing output, this is probably inevitable.

Is `1.1.0` supposed to include the concept of `optional` attributes? If so, including the propsed change for it and `2.X.X` sounds reasonable enough to me. +1

ax3l:
Sorry, I was too short. I overthought what I wrote above and would suggest the following.

I propose to introduce the following relaxation and of the two attributes `meshesPath` and `particlesPath`:

- relax the two attributes from `required` to `optional`
- in the case an attributes is not set, handle the file as if there are no meshes (particles)
[...]

PR #171, opened 2018-01-23
meshesPath & particlesPath: optional

ax3l:
Makes the base attributes `meshesPath` and `particlesPath` optional if no meshes or no particles shall be stored. Specifies that the path they describe MUST exist as soon as they are set. If they are unset, this MUST be interpreted as "no respective record".

*Implements issue:*  #143 (cc @mccoys)

## Description

See [comment in the original issue](https://github.com/openPMD/openPMD-standard/issues/143#issuecomment-346833418):

> I propose to introduce the following relaxation and of the two attributes `meshesPath` and `particlesPath`:
> 
> - relax the two attributes from `required` to `optional`
> - in the case an attributes is not set, handle the file as if there are no meshes (particles)
> - define that the path they describe MUST exist if the attributes are set (as most implemented it already like this in 1.0.0 and our validator does this also in a two-liner)

## Affected Components

- `base`

## Logic Changes

if set, same parser logic as before, but it is now required that the path in the attributes is *valid* (does exist as a group).

If unset, interpret as "no particle/mesh" records found, respectively.

## Writer Changes

No effect for before-valid files.

"Work-arounds" for particle-only/mesh-only openPMD files must now make sure that the path in `meshesPath/` or `particlesPath/` does really exist, since that behavior was unspecified so far. If not needed, they are now allowed to skip an attribute.

## Reader Changes

- [x] `openPMD-validator`: relax validator, check path exists if set: https://github.com/openPMD/openPMD-validator/pull/32
- [ ] `openPMD-viewer`: https://github.com/openPMD/openPMD-viewer/pull/184 https://github.com/openPMD/openPMD-viewer/issues/186
- [x] `yt`: https://github.com/yt-project/yt/pull/1645
- [ ] `openPMD-visit-plugin`: relaxation of reader needed? cc @xxirii @GBlaclard

## Data Converter

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -113,7 +113,7 @@ Each file's *root* group (path `/`) must at least contain the attributes:
       of the form given by `basePath` (e.g. `/extra_data`). In this
       way, the openPMD parsing tools will not parse this additional data. 
 
-The following attributes are *optional* in each each file's *root* directory
+The following attributes are *optional* in each each file's *root* group
 (path `/`) and indicate if a file contains mesh and/or particle records. It is
 *required* to set them if one wants to store mesh and/or particle records.
 
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -113,16 +113,26 @@ Each file's *root* group (path `/`) must at least contain the attributes:
       of the form given by `basePath` (e.g. `/extra_data`). In this
       way, the openPMD parsing tools will not parse this additional data. 
 
+The following attributes are *optional* in each each file's *root* directory
+(path `/`) and indicate if a file contains mesh and/or particle records. It is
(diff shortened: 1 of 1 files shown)


RemiLehe (approved):

RemiLehe:
Great change. I'll implement it in the openPMD-viewer.

DavidSagan:
Looks good.

DavidSagan (approved):
Looks good except "directory" should be "group"

ax3l:
@DavidSagan thx for the review, fixed!

RemiLehe:
Should we merge this now?

ax3l:
all right! :)

Issue #145, opened 2017-11-06
particlePush: none


ax3l:
Allow `none` as `particlePush`, e.g. for static particle such as probes.

RemiLehe:
Sure. I agree with this. Would you like to do a corresponding PR?

ax3l:
It's already open in #172 :)

ax3l:
related to #132

Issue #139, opened 2016-11-28
ED-PIC: Radiation Reaction


ax3l:
Document if radiation reaction effects have been accounted for at each species.

- classical (e.g., DOI:10.1016/j.cpc.2016.04.002 -> into `particlePush`, is also a RK4)
- QED (e.g. DOI:10.1103/PhysRevLett.106.035001, DOI:10.1103/PhysRevSTAB.14.054401, DOI:10.1088/0741-3335/53/1/015009, DOI:10.1103/PhysRevLett.108.165006, DOI:10.1103/PhysRevE.92.023305, can't find a reference to T Grismayer's ~2012/13 implementation besides APS abstracts..., M Lobet 2013/15)

ax3l:
related to #132

Issue #132, opened 2016-07-01
ED-PIC: New Pusher Types


ax3l:
missing pusher types:
- free streaming
- photon pusher? (free streaming with c) -> can be first case with upcoming species type "photon"
- reduced landau lifshitz pusher (RK4 + radiation drag) [arxiv:1502.02432](http://arxiv.org/abs/1502.02432)
- none

PR #172, opened 2018-01-23
ED-PIC: Add Pusher Values

ax3l:
Adds new allowed pusher values for classical radiation reaktion and diagnostics to the `particlePush` attribute in the ED-PIC extensions.

*Implements issues:* #132 #139 #145

## Description

Adds the following new pushers:

- `free-streaming` (constantly moving with initial momentum)
- `LLRK4` (reduced Laundau-Lifshitz pusher via RK4 and classical radiation reaction, [doi:10.1016/j.cpc.2016.04.002](https://dx.doi.org/10.1016/j.cpc.2016.04.002))
- `none` (static particles such as probes)

`none` is a special case of `streaming` but probably still worth to be added to be explicit.

## Affected Components

- `EXT: ED-PIC`

## Logic Changes

None.

## Writer Changes

Writers can now switch from `other` for these three pushers to the newly specified values.

- PIConGPU: https://github.com/ComputationalRadiationPhysics/picongpu/issues/2472

## Reader Changes

Just more allowed values are known now which were previously undefined as `other`.

## Data Converter

Per-application possible but not needed. Old files are still forward compatible to this change.

### DIFF
--- a/EXT_ED-PIC.md
+++ b/EXT_ED-PIC.md
@@ -268,7 +268,12 @@ The individual requirement is given in `scope`.
     - allowed values:
       - `Boris` (J.P. Boris. *Relativistic plasma simulation-optimization of a*
                  *hybrid code.* USA, 1970)
-      - `Vay` ([doi:10.1063/1.2837054](http://dx.doi.org/10.1063/1.2837054))
+      - `Vay` ([doi:10.1063/1.2837054](https://dx.doi.org/10.1063/1.2837054))
+      - `streaming` (constantly moving with initial momentum)
+      - `LLRK4` (reduced Laundau-Lifshitz pusher via RK4 and classical
+                 radiation reaction,
+                 [doi:10.1016/j.cpc.2016.04.002](https://dx.doi.org/10.1016/j.cpc.2016.04.002))
+      - `none` (static particles such as probes)
       - `other`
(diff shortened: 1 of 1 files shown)


RemiLehe:
Instead of `streaming`, I would maybe suggest `ballistic`. Any opinion?

RemiLehe (approved):

ax3l:
Ballistic sounds more physical, I like that.

I just new "free streaming" as a term used in PWFA sims, but I can change it to `ballistic` :)

ax3l:
@RemiLehe thinking about it and reading e.g. the default [wikipedia articles](https://en.wikipedia.org/wiki/Ballistics) on it, yes ballistic means "free to move" but usually under forces such as the gravitational force. I think we want to avoid that impression since e.g. a "free-streaming" electron beam will *not* change its trajectory in background electro-magnetic fields in idealized driver models that use such a pusher.

RemiLehe:
OK, good point. Then maybe we can change `streaming` to `free-streaming`. Does that sound ok?

ax3l:
Good idea, that's more explicit!
or just `free` or `constant` pusher?

RemiLehe:
I think `free-streaming` is the most explicit.

Issue #176, opened 2018-01-30
Standardizing Particle Species Name


DavidSagan:
Standardizing the particle species names will make data files more portable.
I am submitting a pull request with a proposed list of names.

DavidSagan:
@RemiLehe @ax3l: OK a pull request has been submitted.

ax3l:
Offline results from the last VC: the new extension should define an `attribute` that can be put both on mesh and particle records to describe the species type (via it's value), independent of the record name.

The extension should therefore define a new reserved `record` attribute and allowed `values` for it.

DavidSagan:
@ax3l: OK how about `SpeciesName` or `SpeciesType` for the attribute name?

ax3l:
sounds good :)

ax3l:
or `PhysicalSpecies`?

Issue #163, opened 2017-12-10
Cross-Record Grouping


DavidSagan:
Right now there is only one root group for meshes: `/data/meshes/`. As opposed to particles where where there can be many: `/data/particles/electrons/`, `/data/particles/protons/`, etc.

This is restricting. For example it may be desirable to store internal fields separately from external fields. So to have a structure to enable groups like: `/data/meshes/dc_field/`, /data/meshes/particle_field/`, would be desirable.

@ax3l: Please put this in the 2.0 project.

ax3l:
@DavidSagan thank you for the proposal!

Can you elaborate a little on what is missing? Can't this be realized with individual names of meshes for external fields already as in your example?

Why would we need to group the fields further instead of just naming them as a domain-expert would understand?

DavidSagan:
@ax3l 

Looking at this issue again I might be confusing the base standard with the ED-PIC extension. In any case, if I look at the example openPMD datasets I see paths like `/data/400/fields/E` and there is no way to specify multiple E-fields since `meshesPath` is `fields`. My proposal is to add a directory level so that you could have `/data/400/fields/dc-field/E`, `/data/400/fields/ac-field/E`, etc.

> Why would we need to group the fields further instead of just naming them as a domain-expert would understand?

I don't understand this.

RemiLehe:
@DavidSagan I think that Axel's point is that you can always use:
` /data/400/fields/Efield_dc` and `/data/400/fields/Efield_ac` (which would be compatible with the current standard) instead of `/data/400/fields/dc-field/E` and `/data/400/fields/ac-field/E`.

Would that be okay with you?

DavidSagan:
@RemiLehe 

> I think that Axel's point is that you can always use:
/data/400/fields/Efield_dc and /data/400/fields/Efield_ac (which would be compatible with the current standard) instead of /data/400/fields/dc-field/E and /data/400/fields/ac-field/E.

If I look at "Naming Conventions for mesh records (field records)" section of the EXT_ED-PIC extension it looks like `/data/400/fields/Efield_dc` would not be compatible since, in this example,  `meshesPath` is set to "fields/". But this is an extension standard and it looks like the base standard does not impose this so all this may just be my confusing the extension with the base.

ax3l:
[...]

PR #177, opened 2018-01-30
Added ParticleSpeciesNames.md

DavidSagan:
This pull request has a new file for standardizing the names of particle species. See #176 .

Affected Components: None.
Writer/Reader/Data converter Changes: At this point none since the present standard does not reference particle species names.


ParticleSpeciesNames.md
```diff
@@ -0,0 +1,47 @@
...
+  - up-antiquark
+  - up-quark
+  - w-boson
+  - z-boson
+
+
+Names for Atoms and Molecules
+-------------------
+
+ - Atoms and molecules: Use standard chemical notation. Eg: "H20". Isotopes are denoted by a pound symbol "#" followed by the isotopic number followed by the chemical symbol. Eg: "#3He" for Helium-3.
```
ax3l: maybe that's a stupid question, but is there some kind of ISO naming already for this on which we can rely on?

ax3l (commented):
We will need to add:
- [ ] an extension ID `4` for 1.1.0 (because it will first be removed in 2.0.0)
- [ ] mention it in the base standard in the extensions section
- [ ] add the name  for the new `record` attribute
- [ ] define if this attribute is optional/recommended/required if the extension is declared for a file
- [ ] define if it is allowed to have more than one particle name per record, e.g. one could store a mesh with "summed charge density" of all ions and electrons -> list?

ParticleSpeciesNames.md
DavidSagan: I looked but I could not find anything unfortunately.

ax3l: I see, thanks for checking.
And something from PDG (the particle data group), maybe? Just that we don't miss something that already defines spellings.

ax3l: (besides that insecurity of mine about accidentally re-inventing something, I like the proposed spellings)

DavidSagan: I had looked at the PDG as well without luck.

DavidSagan:
I realize I forgot to specify how to write atom and molecular charge states. So how about something like:

The charge state is denoted by a suffix with either:
1. Some number of plus or minus signs. Example: "H2O++"
1. A plus or minus sign with a following integer. Example: "CO+3"

ax3l:
I wonder if we want to record charge states really in the species type name.

We usually have them as a record in the species anyway and for constant charge state as a constant record. Maybe other domains would do the same?

DavidSagan:
@ax3l: Having the charge state separately specified would be fine.

ax3l:
Changes I applied:
- file renamed to `EXT_SpeciesType.md`
- extension name: `SpeciesType` with extension-ID `4` (`1` is used and we reserved `2` in the past)
- formatting unified
- defined `SpeciesType` as attribute: *recommended* if extension is declared
- allowed `lists`, e.g. for a local *summed charge density* in a non-equilibrium plasma
- atoms & isotopes: reference to IUPAC namings
- reference in base standard

EXT_SpeciesType.md
```diff
@@ -0,0 +1,80 @@
...
+
+Additional Record Attribute
+---------------------------
+
+The following additional attribute for openPMD `records` is defined in this
+extension:
+
+- `SpeciesType`
+  - type: *(string)*
+  - scope: *recommended*
```
ax3l: offline discussion: maybe better *optional* here, it's weird for e.g. "E fields" or arbitrary other records

Later on, in the `ED-PIC` extension or the `BeamPhysics` extension one can still specify, that it is *recommended* for e.g. `particle species`.

EXT_SpeciesType.md
```diff
@@ -0,0 +1,80 @@
...
+  - description: particle species in this record, semicolon-separated list
```
DavidSagan: Replace "semicolon-separated list" with "If there are multiple species to be specified, they can be specified using a semicolon separated list."

EXT_SpeciesType.md
```diff
@@ -0,0 +1,80 @@
...
+  - examples:
+    - `electron` (e.g. on an electron `particle record` or an electron
+                  density `mesh record`)
+    - `electron;proton;#12C` (e.g. on a `mesh record` for a plasma's
+                              local charge density)
+    - `other:apple;other:orange` (for a `record` mixing apples & oranges)
+
+This attribute can be used with any `record` (including `mesh records`).
+
+### SubAtomic Particle Names
```
ax3l: Unify headline: SubAtomic Particles

EXT_SpeciesType.md
```diff
@@ -0,0 +1,80 @@
...
+  - `top-antiquark`
+  - `top-quark`
+  - `up-antiquark`
+  - `up-quark`
+  - `w-boson`
+  - `z-boson`
+
+### Atoms & Isotopes
+
+Isotopes are denoted by a pound symbol `#` followed by the isotopic number followed by the chemical symbol, e.g.: `#3He` for Helium-3.
```
DavidSagan: Switch order sentences. That is, start this section with "Element namings...".

EXT_SpeciesType.md
```diff
@@ -0,0 +1,80 @@
...
+Element namings follow the abbreviated namings of the periodic table, defined by *The International Union of Pure and Applied Chemistry* (IUPAC).
+
```
ax3l: Should we mention that the most abundant isotope can just be short-handed as `He` (for `#4He`)? It's somehow missing, but would allow ambiguity.

Is `D` ok for `#2H` or shall it be forced to the latter?

DavidSagan: I would say that the standard should assume that  "He" is equivalent to "#4He".

DavidSagan: Deuterium would be "#2H"

ax3l: excellent, so we keep the abbreviations to a minimum: regular elements in the IUPAC periodic table

EXT_SpeciesType.md
```diff
@@ -0,0 +1,80 @@
...
+### Molecules
+
+Use standard chemical notation, e.g.: `H20`.
```
DavidSagan: The isotope prefix can be used with molecules. Example "#2H2O"  for two deuterium and "#2HHO" for one deuterium.

ax3l: idea: we prefix the allowed values with the sections, e.g. `subatomic:`, `atom:`/`isotope:`, `molecule:`, ...

EXT_SpeciesType.md
```diff
@@ -0,0 +1,90 @@
+Convention for Specifying Particle Species 
+==========================================
+
+openPMD extension name: `SpeciesType`
+
+openPMD extension ID: `4`
```
ax3l, line 6: note: this will be removed in `2.0.0` but we have to assign it since we are still in `1.X`

EXT_SpeciesType.md
```diff
@@ -0,0 +1,90 @@
...
+                  density `mesh record`)
+    - `electron;proton;#12C` (e.g. on a `mesh record` for a plasma's
+                              local charge density)
+    - `other:apple;other:orange` (for a `record` mixing apples & oranges)
+
+This attribute can be used with any `record` (including `mesh records`).
+
+### SubAtomic Particles
+
+  - `subatomic:antimuon`
```
ax3l: offline discussion: we won't collide (yet) and if we should later one we can still add prefixes and convert known ones.

ax3l:
@RemiLehe we finished our proposal and you can now add your comments & review! :)

We tried to document all ideas and steps during our review today inline in comments, so if you like you can just un-collapse them for curiosity.

ax3l:
@RemiLehe ping :)

EXT_SpeciesType.md
```diff
@@ -0,0 +1,94 @@
...
+  - `neutron`
+  - `photon`
+  - `pion`
+  - `positron`
+  - `proton`
+  - `strange-antiquark`
+  - `strange-quark`
+  - `tao-antiquark`
+  - `tao-quark`
+  - `tao-neutrino`
```
RemiLehe: Shouldn't this be `tau-neutrino` (and adding `tau-antineutrino`) instead of `tao-neutrino` ? In addition, we should probably add `tau` as a particle (for the counterpart of `muon` and `electron`).

EXT_SpeciesType.md
```diff
@@ -0,0 +1,94 @@
...
+  - `electron`
+  - `electron-neutrino`
+  - `gluon`
+  - `graviton`
+  - `higgs-boson`
+  - `muon`
+  - `muon-neutrino`
+  - `neutron`
+  - `photon`
+  - `pion`
```
RemiLehe: Is there a reason to standardize the name of the `pion`, but not of all the other mesons?
(If the reason is that the `pion` is the most common, I am fine with this ; I am definitely not a specialist... I was just wondering.)

EXT_SpeciesType.md
```diff
@@ -0,0 +1,94 @@
...
+                  density `mesh record`)
+    - `electron;proton;#12C` (e.g. on a `mesh record` for a plasma's
+                              local charge density)
+    - `other:apple;other:orange` (for a `record` mixing apples & oranges)
+
+This attribute can be used with any `record` (including `mesh records`).
+
+### SubAtomic Particles
+
+  - `antimuon`
```
RemiLehe: Should we add `antitau`? (See remark below on the `tau` lepton.)

EXT_SpeciesType.md
```diff
@@ -0,0 +1,94 @@
...
+  - `antiproton`
```
RemiLehe: Should we add `antineutron`?

EXT_SpeciesType.md
```diff
@@ -0,0 +1,94 @@
...
+  - `bottom`
```
RemiLehe: Shouldn't this be `bottom-quark` and `bottom-antiquark` for consistency?

EXT_SpeciesType.md
```diff
@@ -0,0 +1,94 @@
...
+  - `charm-antiquark`
+  - `charm-quark`
+  - `deuteron`
```
RemiLehe: It seems that `deuteron` could be expressed as `#2H`. Using the convention below. Therefore, I suggest that we remove it from  this list.

EXT_SpeciesType.md
```diff
@@ -0,0 +1,94 @@
...
+  - `muon-neutrino`
+  - `neutron`
+  - `photon`
+  - `pion`
+  - `positron`
+  - `proton`
+  - `strange-antiquark`
+  - `strange-quark`
+  - `tao-antiquark`
+  - `tao-quark`
```
RemiLehe: Also, I am not sure whether a `tao-quark` and `tao-antiquark` exist in the standard model... (If I remember well, the quarks of the standard model are `up`, `down`, `strange`, `charm`, `bottom`, `top`.)

EXT_SpeciesType.md
```diff
@@ -0,0 +1,94 @@
...
+
+Element namings follow the abbreviated namings of the periodic table, defined
+by *The International Union of Pure and Applied Chemistry* (IUPAC).
+An example would be `Si` for silicon.
+
+Specifications of isotopes are denoted by a pound symbol `#` followed
+by the isotopic number followed by the chemical symbol, e.g.: `#3He`
+for Helium-3.
+
+The charge state is not encoded by the `SpeciesType` attribute.
```
RemiLehe, line 86: By `charge state`, do we mean `ionization state`? Could `charge state` be considered ambiguous for some users?

RemiLehe (changes requested):

RemiLehe:
Sorry for my late review! I added a couple of comments/questions.

EXT_SpeciesType.md
```diff
@@ -0,0 +1,94 @@
...
+  - `antimuon`
+  - `antiproton`
+  - `bottom`
+  - `charm-antiquark`
+  - `charm-quark`
+  - `deuteron`
+  - `down-antiquark`
+  - `down-quark`
+  - `electron`
+  - `electron-neutrino`
```
RemiLehe: Should we add `electron-antineutrino`?

EXT_SpeciesType.md
```diff
@@ -0,0 +1,94 @@
...
+  - `gluon`
+  - `graviton`
+  - `higgs-boson`
+  - `muon`
+  - `muon-neutrino`
```
RemiLehe: Should we add `muon_antineutrino`?

RemiLehe (changes requested):

EXT_SpeciesType.md
ax3l: Imho, charge state is the more general concept while ionization state only applies to atoms & molecules. This section just says we do not express it in this extension.

RemiLehe: OK, sounds good.

EXT_SpeciesType.md
DavidSagan: Yes we should add `antitau`. And `tao` is a typo. Should be `tau` everywhere.

EXT_SpeciesType.md
DavidSagan: Yes.

EXT_SpeciesType.md
DavidSagan: Yes.

EXT_SpeciesType.md
DavidSagan: Correct that `deuteron` is the same as `#2H` with a charge state of 1. In terms of having the name in the standard I don't care either way. @ax3l: Do you have an opinion?

EXT_SpeciesType.md
DavidSagan: The list I came up with was meant to get the discussion going rather then  being complete (I am definitely not a specialist either). Do you have a list of names for the other mesons? By all means let us put them in.

EXT_SpeciesType.md
DavidSagan: Correct. That is a typo and should be removed.

EXT_SpeciesType.md
DavidSagan: Yes. The change should be made.

EXT_SpeciesType.md
DavidSagan: I concur with Alex.

EXT_SpeciesType.md
DavidSagan: Sure.

EXT_SpeciesType.md
DavidSagan: Yes.

EXT_SpeciesType.md
ax3l: sure, Deuterium should not be here

ax3l:
Sorry, I totally forgot to swipe through the subatomic particle list again in review.

I would suggest to make one list with elementary particles of the standard model (and their anti-partners), Gauge bosons and the higgs boson.

There is a plethora of namings for composite particles that are only used in elementary particle physics, especially Hadrons such as Mesons and Baryons, and should imho not be tried to be defined here (besides the obvious ones we need such below for atoms), but instead reference e.g. the [particle data group](http://pdg.lbl.gov/2017/tables/contents_tables.html). Also, actual particle detector people think of jets and channels with particles which are again composites of the above. As far as we have no detector people here that suggest a naming I would also keep this free. Likely they would also use group identifiers #163

EXT_SpeciesType.md
ax3l: And I thought you wanted to point to Majorana neutrinos here ;)

DavidSagan:
@ax3l 

>I would suggest to make one list with elementary particles of the standard model (and their anti-partners), Gauge bosons and the higgs boson.

Sounds good to me.

ax3l:
I updated the PR accordingly.

I would suggest to move this PR to `2.0.0` and close the `1.1.0` merge window so we do not merge it in last-minute and can take more time for it. Is that ok? I would then prepare the last clean-ups for `1.1.0` now.

EXT_SpeciesType.md
```diff
@@ -0,0 +1,95 @@
...
+  - `up` (`anti-up`)
+  - `down` (`anti-down`)
+  - `charm` (`anti-charm`)
+  - `strange` (`anti-strange`)
+  - `top` (`anti-top`)
+  - `bottom` (`anti-bottom`)
+
+Leptons:
+  - `electron` (`positron`)
+  - `electron-neutrino` (`anti-electron-neutrino`)
```
ax3l, line 55: anti-neutrinos: who knows, maybe they turn out to be the same particles. but until now it seems that calling them anti-neutrinos is common

DavidSagan:
@ax3l: 
>I would suggest to move this PR to 2.0.0 and close the 1.1.0 merge window so we do not merge it in last-minute and can take more time for it. Is that ok? 

Fine with me.

RemiLehe (approved):

RemiLehe:
I really like the reorganized version of the file. 
Thanks for editing the PR!

DavidSagan:
Yes I like the reorg too.
Minor stylistic suggestion: Use a comma instead of parenthesis. For example, Instead of:
  `electron ( positron )`
write:
  `electron, positron`

DavidSagan:
I know we don't want to in general get into trying to define names of mesons but thinking about it I would make an exception for `proton` and `anti-proton` since these are commonly used in experiments.

EXT_SpeciesType.md
ax3l: *Axel ;-)

ax3l:
I am ok with both suggestions, just need to find a place to put `anti-proton` and maybe to say that `H` and `proton` are the same thing? Actually they might not be the same, a user might want to refer to the *element* (atom) `H`...

DavidSagan:
>I am ok with both suggestions, just need to find a place to put anti-proton and maybe to say that H and proton are the same thing? Actually they might not be the same, a user might want to refer to the element (atom) H...

It could be added that `#1H` is the same as `proton` if the charge is specified as `+1`.

ax3l:
I might just have been thinking loudly. `proton` is just more specific than `H` and that is fine, imho. No reason to explain more I think.

ax3l:
we will re-open this PR for further review into `upcoming-2.0.0`

ax3l:
Re-opened for 2.0.0 inclusion in #180

PR #180, opened 2018-02-06
New Extension: SpeciesType

DavidSagan:
This pull request has a new file for standardizing the names of particle species.

*Implements issue:* #176

## Description

This convention is for standardizing the names of particle species, e.g. in particle physics.

## Affected Components

- new extension: `SpeciesType`

## Logic Changes

A new *optional* string attribute `speciesType` can now be present at `mesh record` and `particle group` level.

## Writer Changes

- [ ] `openPMD-api` (upcoming): Add to writer and reader API.

## Reader Changes

Readers can if they want read the new attribute (if it exists) and propose e.g. groups, accumulations, etc. on it.

- [x] `openPMD-validator`: Validate it's a string if present https://github.com/openPMD/openPMD-validator/pull/38
- [ ] `openPMD-viewer`: https://github.com/openPMD/openPMD-viewer/...

## Data Updater

New optional extension with optional attribute. No update of existing data needed.
Edit: update needed for `ED-PIC` extension: #192


ax3l:
Left-over to do's from #177:

- [x] hardons: define protons & anti-protons
- [x] separate anti-particles by `,` instead of putting them in brackets
- [x] remove extension-ID (2.0.0)

DavidSagan:
@ax3l There also should be a version number associated with this. I assume it should be 1.0.0 .

ax3l:
@DavidSagan the version number will simply be the one of the standard it gets in with.

ax3l (approved):

DavidSagan:
@ax3l 

> the version number will simply be the one of the standard it gets in with.

OK just to make sure I understand: Say in the future changes to an extension necessitates that it is to be bumped up a version. Then even if the base standard has not changed it will be bumped up too? Yes?

ax3l:
Exactly.
Admittedly, up to now the major changes were triggered by the base standard.

ax3l:
@RemiLehe @marcguetg hi, do you want to give your review comments on this PR? From my side I think it's already in quite good shape.

EXT_SpeciesType.md
```diff
@@ -0,0 +1,98 @@
...
+------------
+
+This convention is for standardizing the names of particle species, e.g. in
+particle physics.
+
+
+Additional Record Attribute
+---------------------------
+
+The following additional attribute for openPMD `records` is defined in this
```
ax3l: oh wait, I think it should read:

... *mesh `records`* and *particle* `groups` ...

DavidSagan: How about simply:

The following attribute for openPMD is defined ...

ax3l: Sure, but I think for reader tools it is good to know where to look for it, no?
I currently only see these two locations as useful and relaxing later on is always possible.

DavidSagan: OK I did not realize that this was defining where the info should go. How about adding a definition for `mesh record` and `particle group` in the Standard?

RemiLehe (approved):

RemiLehe:
@ax3l Should we merge this?

EXT_SpeciesType.md
ax3l: good, this aspect is moved to #187

ax3l:
ok, I addressed the last inline comment about the location of the `SpeciesType` attribute, so reader implementors know where to expect it

DavidSagan:
Minor detail: Since `neutron` is a common standard name do we want to use `neutron`, `anti-neutron` rather than `other:neutron`, `other:anti-neutron`?

ax3l:
Sure, we can do that. It's just not too common for us since it's neutral ;-)

I did just draw a line (arbitrarily) since (anti-)protons are the only stable free hadrons.

ax3l:
@DavidSagan moved (anti-)neutron to "common" and added sigma baryon as another example. now we draw the line at the nucleus :D

EXT_SpeciesType.md
```diff
@@ -0,0 +1,99 @@
...
+particle physics.
+
+
+Additional Record Attribute
+---------------------------
+
+The following additional attribute for openPMD `mesh records` and
+`particle groups` is defined in this extension:
+
+- `SpeciesType`
```
ax3l: wait, why is this upper-case in the first letter? Aren't we starting all attributes with a lowercase letter? Will update for consistency.

ax3l (changes requested):

ax3l (approved):

ax3l:
@RemiLehe @DavidSagan I think this can be merged now :)

EXT_SpeciesType.md
```diff
@@ -0,0 +1,99 @@
...
+  - `other:kaon`, ...
+
+### Atoms & Isotopes
+
+Element namings follow the abbreviated namings of the periodic table, defined
+by *The International Union of Pure and Applied Chemistry* (IUPAC).
+An example would be `Si` for silicon.
+
+Specifications of isotopes are denoted by a pound symbol `#` followed
+by the isotopic number followed by the chemical symbol, e.g.: `#3He`
```
zerothi, line 87: Could we add a mention on how to mention `#2H2#15O`? It isn't fully clear how this works in this flow?

Issue #182, opened 2018-02-15
Rename position and positionOffset


DavidSagan:
Right now there is `position` and `positionOffset` under "Records for each Particle Species". The names here are a bit confusing so I propose renaming these to:
```
  position             -> positionOrigin
  positionOffset       -> position
```
And `positionOrigin` would only be implemented with constant record components.

DavidSagan:
Another issue: positions with the Beam Physics extension can be phase space coordinates where the **z** coordinate is not actually a position. My thought is to make all position records in the base standard optional and then the Beam Physics extension can appropriately supply the necessary position definitions.

ax3l:
The two-part separation is an implementation for particle-mesh codes (the PM in openPMD), potentially AMR codes and codes dealing with neighbor lists.

Some particle-mesh codes, such as PIConGPU, store positions on two levels: *coarse* up to the nearest associated grid point and *fine* within two grid points. This has several advantages for large-scale simulations, mainly that floating precision does not change over the simulation box.

> And `position[Offset]` would only be implemented with constant record components.

This is only in the Warp example files the case. Our data sets really store two, additive, non-constant record components.

Naming this offset origin would be misleading in our case, it's really used for two constituents of the position: an "offset" and a "fine" position.

DavidSagan:
@ax3l OK. The Beam Physics extension is likely to have a different organization so how about making these optional?

ax3l:
VC result: we will read the new proposed `BeamPhysics` extension PR and come back to this question of unifying "origin"/"offset" namings.

ax3l:
We realized in the last VC this issue is probably outdated. I will add a label for now, @DavidSagan feel free to comment or close it as applicable :)

DavidSagan:
@ax3l  No it is not outdated. In fact there is a conflict now with the BeamPhysics extension that needs to be resolved.

ax3l:
@DavidSagan no problem, moved back on the agenda! :)

ax3l:
VC offline result today:
there were two questions in this issue regarding the base standard's `position` particle record.

a) "Is `position` as a name ok or confusing in a generalized phase space position?" In the `BeamPhysics` extension, it will have a longitudinal component that can have various meanings and names (but is still a length). -> Yes, still ok. We don't require strict frames of reference of the base standard's `position` and the *record components* can be arbitrarily named.

[...]

PR #183, opened 2018-02-20
New Extension: BeamPhysics

DavidSagan:
This is the new proposed `Beam Physics` extension for describing particle beams in accelerators, storage rings, etc.

CC @ChristopherMayes


EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
+Extension to the openPMD Standard for Describing Particle Beams and X-rays
+==========================================
+
+Version 1.0.0 (March 1, 2018)
```
RemiLehe: As discussed off-line, we should remove the version number here.
I think that the convention we decided is that modifying one of the extension would bump the version number of the whole standard.

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+
+- **Global coordinate system**: The **global** coordinate system is a coordinate system that is
+used to describe the position and orientation of machine elements in space. That is, the **global**
+coordinate system is fixed with respect to the building or room where the machine is placed independent of the
+machine itself.
+
+- **Lattice coordinate system**: The curvilinear coordinate system whose "longitudinal"
+coordinate (typically called **s**) typically runs through the nominal centers of the elements
+in the machine. Typically, the **lattice** coordinate system is used to describe misalignments
+of lattice elements with respect to the rest of the lattice.
```
RemiLehe: This is really minor, but the word "typically" occurs 3 times in this paragraph. Maybe this could be rephrased?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+
+- **Group**: A **group** is a container structure containing
+a set of zero or more **attributes**, a set of zero or more **groups** (which can be called
+**sub-groups**), and a set of zero or more **datasets**. Note: In HDF5, these are also
+called **groups**.
+
+- **Dataset**: A **dataset** is a structure that contains a set of zero or more **attributes** and a
+data aray (which may be multidimensional).  Note: In HDF5, these are also called **datasets**.
+
+- **Record**: A **record** is a **group** (without any sub-groups) or a **dataset** that contains data
```
RemiLehe: Would you agree to add "As explained in the base standard," or "As a reminder from the base standard" at the beginning of the paragraph? This is so as to avoid confusion for readers that might think  that this redefines `record`.

In fact, I am wondering if it is necessary to reiterate those definitions here. Ideally, things should be defined only once. If the issue is that those definitions were not clear enough in the base standard, maybe we should rephrase the base standard itself. @ax3l : Any opinion?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+    **record** is a **group**. If the particles have differing charges, the values are stored in an array
+    of the **scalar record**. In this case the **scalar record** is a **dataset**.
+    - **vector records** hold a set of **datasets**. In this case the **record** is a **group**.
+        - Example: A **record** named **E** for holding electric field values may have three datasets holding
+        the components of the field named **E/x**, **E/y**, and **E/z**.
+
+- **Attribute**: An **attribute** is a variable associated with a **group** along with a
+value. Example: **snapshotPath** is a string variable associated with the root **/** group.
+
+- **Polar coordinates**: **Polar** coordinates are **(r, theta, phi)** where **r** is the radius, **theta** is he angle
```
RemiLehe: "he angle" should be replaced by "the angle"

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+from the **z** or **s** axis, and **phi** is the projected azimuthal angle in the **(x, y)**
+plane.
+
+- **Particle Root Group**: The **Particle Root Group** is the root group for specifying a group of particles. There can be multiple **particle root groups** in a data file. For example, a set of particle bunches might specify one particle root group for each bunch. See the Base Standard for more information.
```
RemiLehe: I am slightly confused by this. By "particle root group" do you mean the group to which `particlesPath` points, or do you mean the sub-groups thereof?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+
+Notes:
+------
+
+- When using the **lattice** coordinate system, the `position` coordinates are **(x, y, s)** where
+nominally **x** is the "radial" component,
```
RemiLehe: This seems to imply a ring-type of accelerator. Would this standard also apply for linear accelerators? If yes, could we add a note here that for linear accelerators, "x" and "y" can be arbitrary?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+
+The following attributes are defined for the file root group.
+
+- `latticeName`
+  - Type: Optional *(string)*
+  - Description: The name of the lattice.
+
+- `latticeFile`
+  - Type: Optional *(string)*
+  - Description: The location of the root lattice file.
```
RemiLehe: Interesting: So far in the openPMD standard we never encountered the case in which the data needs another file. 
One issue is that, if the openPMD file (+ the complementary lattice file) is moved to another machine for postprocessing, then this path will not be valid anymore. I am not sure how to solve this issue. 
Maybe this falls in the category of trying to store to much information in the openPMD file. For instance, for PIC simulation of laser-plasma interaction, we decided not to store the physical parameters of the setup (e.g. profile of the laser, profile of the plasma, etc.) because it would be too complicated. Any opinion?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+branch index integer, and **element-index** is the associated lattice element index within the branch.
+
+- `locationInElement`
+  - Type Optional *(int)*
+  - Description: This attribute is used with  `latticeElementID`/`latticeElementName` to specify
+  the origin where the particle or particles are measured with respect in the **lattice** coordinate system.   
+  - The origin is always on the longitudinal axis (x = y = 0) of the **lattice** coordinate system.
+  - Possible values:    
+    - `Upstream-End`: Upstream end of element outside of any edge fields.
+    - `<s-position>`: Where `<s-position>` is a number. Inside the element at a distance, given by `<s-position>`,
```
RemiLehe: This paragraph specifies that `locationInElement` is an *int*. But from the possible values it seems that it could be a string. Could you clarify this point?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+  - Note: Since some programs will model edge fields of a lattice element as having zero length, the longitudinal **s**-position
+does not necessarily provide enough information to determine where a particle is with respect to an edge field so `Upstream-End` and `Downstream-End` are provided.
+
+- `momentumNormalization`
+    - Type: Optional *(string)*
+    - Description: Normalization used for momentum values.
+    - Possible values:
+        - `referenceTotalMomentum`: Normalize with respect to the `referenceTotalMomentum` attribute
+        - `referenceEnergy`: Normalize with respect to the `referenceEenergy` attribute.
+        - `none`: No normalization.
```
RemiLehe: `referenceEenergy` should probably be replace by `referenceEnergy`.

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+
+- `referenceEnergy`
+    - Type: Optional *(float)* attribute.
+    - Description: Specifies the reference energy from which the `energy` is measured with respect to.
```
RemiLehe: Should we impose the unit of this attribute?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+
+- `referenceTotalMomentum`
+    - Type: Optional *(float)* attribute
+    - Description: Specifies the reference total momentum from which the total momentum is measured with respect to.
```
RemiLehe: Should we impose the unit of this attribute?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+    - Description: The total charge of all the particles.
+
+Per-Particle Records of the `Particle Root Group`
+---------------------
+
+The following records store data on a particle by particle basis.
+
+- `time-refTime`
+- Type: Optional *(float)*
+- Description: Particle time minus the referece time.
```
RemiLehe: `referece` should probably by replaced by `reference`

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+
+- `energy/`
```
RemiLehe: There is probably no need for the trailing `/` here. Is that correct?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+  - Type: Optional *(float)*
+  - Description: The total energy of the particles relative to the `referenceEnergy` attribute.
```
RemiLehe: Could you add a quick math formula to explain how the total energy can be retrieved from the `referenceEnergy`?
Also, does this mean that `referenceEnergy` is *not* optional whenever `energy` is present?
Finally, since `referenceEnergy` seems to be simply a normalization/conversion factor, I am wondering if it is not redundent with the `unitSI` coefficient...

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+
+- `electricField/`
+    - Type: Optional 2-component *(float)*
+    - Description: Electric field. Used for photons only.
```
RemiLehe: Out of curiosity: what do you mean by "Used for photons only" ?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+    - Components: (`x`, `y`).
+        - For each component, the field is specified using either (`amplitude`, `phase`) or (`Real`, `Imaginary`)
+        subcomponents.
+
+- `macroCharge/`
+    - Type: Optional *(float)*
+    - Description: "Macro"-particles are simulation particles that represent multiple real particles.
+    The `macroCharge` is the the total charge of the collection of particles that a macro-particle
+    represents.
```
RemiLehe: This seems similar to `weighting` (as well as the attribute `weightingPower`) in the [ED-PIC extension](https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/EXT_ED-PIC.md). 

Should we maybe use the same conventions (and maybe isolate this in an extension specifically for macroparticles)?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+
+- `momentum/`
+    - Type: Optional 2 or 3-vector *(float)*
+    - Description: The total momentum of the particles relative to the `momentumOrigin` attribute.
```
RemiLehe: Note that, unless I missed it, the `momentumOrigin` attribute is not defined in the present document.

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+    - Components: (`px`, `py`) or (`px`, `py`, `pz`).
+    - Note: A program using phase space coordinates to describe particle positions will typically use the transverse momentum
+    `px` and `py` along with the `totalMomentum` or `energy` records.
+
+- `pathLength/`
```
RemiLehe, line 188: The trailing slash is probably not needed here.

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+    - Type: Optional *(float)*
+    - Description: Length that a particle has traveled.
+
+- `position`
+    - Type: Required 3-vector *(float)*
+    - Description: Position relative to the coordinate origin.
+    - Components: (`x`, `y`, `<z-coord>`) where `<z-coord>` is one of:
+        - `z`: True longitudinal coordinate.
```
RemiLehe: Is `z` in the lattice coordinates? If yes, could a user also use the letter `s` in the openPMD file?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+        - Where: **beta** = particle speed, **c** = speed of light, **t** = time, and **dt** = time relative to the reference time.
+
+- `refTime/`
+    - Type: Optional *(float)*
+    - Description: The reference particle time.
+    Note that the reference time is a function of **s** and therefore can be different for different particles.
+
+- `s-Position`
+    - Type: Optional *(float)*
+    - Description: Absolute longitudinal distance in **lattice** coordinates.
```
RemiLehe: Could you explain the difference with the `z` position above?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+- `speed/`
+    - Type: Optional *(float)*
+    - Description: The speed of the particles.
+
+- `spin/`
+    - Type: Optional 3-vector *(float)*
+    - Description: Particle spin.
+    - Components: (`x`, `y`, `s`) or (`r`, `theta`, `phi`).
+
+- `time/`
```
RemiLehe, line 208: This seems redundant with `time-refTime`.
Could you we instead always use `time-refTime`, and set `refTime` as a constant record equal to 0?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+    - Type: Optional *(float)*
+    - Description: Absolute particle time. Note: Particles may have different times if the snapshot
+    is, for example, taken at constant **s**.
+
+- `totalMomentum/`
+    - Type: Optional *(float)*
+    - Description: The total momentum of the particles relative to the `referenceTotalMomentum` attribute.
```
RemiLehe: Just for my own understaning, does this mean `sqrt(px**2 + py**2 + pz**2)`?
Also, the same remarks as for `energy` apply (regarding normalization and redundancy with `unitSI`).

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+
+Non Per-Particle Records of the `Particle Root Group`
+---------------------
+
+The following possible records of the `Particle Root Group` are for specifying properties of the entire group of particles.
+
+- `phaseSpaceFirstOrderMoment/`
+  - Type: Optional 6-vector *(float)*
+  - Description: First order beam moments.
```
RemiLehe: Should we specify the order of the 6 vector (`<x>, <y>, <s>, <px>, <py>, <pz>`) for instance.
If there is a very standard convention in lattice/accelerator physics, maybe we should just mention that this use that convention.

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,269 @@
...
+
+- `phaseSpaceSecondOrderMoment/`
+  - Type: Optional 6x6-matrix *(float)*
+  - Description: Second order beam moments.
+
+- `phaseSpaceThirdOrderMoment/`
+  - Type: Optional 6x6x6-tensor *(float)*
+  - Description: Third order beam moments.
```
RemiLehe: Same questions on the order of the 6 components.

EXT_BeamPhysics.md
DavidSagan: Agreed

EXT_BeamPhysics.md
DavidSagan: How about:

- **Lattice coordinate system**: The curvilinear coordinate system whose "longitudinal"
coordinate (**s**) typically runs through the nominal centers of the elements
in the machine. The **lattice** coordinate system is often used to describe particle positions.

EXT_BeamPhysics.md
DavidSagan: @RemiLehe 

I agree that definitions should only appear once so this should be revised. "Data set" is not defined in the base Standard so I would propose just moving the definition here to the base Standard. [By the way, I use "dataset" rather than "data set". I don't care what the convention is but we should use one or the other but not both.]

Also "record" in the base Standard is used before it's definition. I think it would be good to have a definitions section in the base Standard near the top of the file so that words are defined before they are used.

EXT_BeamPhysics.md
DavidSagan: Agree.

EXT_BeamPhysics.md
DavidSagan: sub-groups thereof. How about changing this to:

- **Particle Root Group**: A **Particle Root Group** is a group for specifying a set of particles. There can be multiple **particle root groups** in a data file. **Particle root groups** are subgroups of the `particlesPath` group defined in the Base Standard.

?

EXT_BeamPhysics.md
DavidSagan: **s**-position is the distance along the **s**-axis of the lattice coordinate system. 

**z** is the 5th component of the particle's phase space position. The definition of **z** is different for different programs. For example, with my Bmad software, **z** is defined to be -beta*c*(t_ref-t) where beta is the particle speed, c is the speed of light, t_ref is the time associated with the reference particle, and t is the time associated with the particle under consideration. The reason for the funny definition is to have canonical coordinates. 

One thing to keep in mind is that there is no simple relationship between **s** and **z**.

EXT_BeamPhysics.md
DavidSagan: `time-refTime` is typically orders of magnitude smaller than `time` so you needed to specify both to avoid roundoff. We could eliminate `refTime` if needed. It typically is different for different particles so a constant record will generally not be suitable.

EXT_BeamPhysics.md
DavidSagan: @RemiLehe 

> Just for my own understanding, does this mean sqrt(px**2 + py**2 + pz**2)?

Yes if px, py, pz are unnormalized momentum coordinates. Typically, px, and py used by accelerator programs are normalized momentum coordinates and pz could be something like energy deviation and not at all directly related to the longitudinal momentum.

EXT_BeamPhysics.md
DavidSagan: Typically the order is (x, px, y, py, z, pz). I cannot recall any program using (x, y, z, px, py, pz). 

I think it would be good to mention this as you suggest.

EXT_BeamPhysics.md
DavidSagan: Yes it should be mentioned that (x, px, y, py, z, pz) should be used.

EXT_BeamPhysics.md
DavidSagan: OK let's change "radial" to "horizontal".

EXT_BeamPhysics.md
DavidSagan: The lattice file is not needed for, say, viewing the data so in that sense there is no **direct** dependency. However, if one wanted to propagate the particles further one would need an appropriate lattice file. 

If there were many attributes used here to store lattice information I would agree with you about trying to store too much. However this is only one attribute and there will be circumstances where this information will be very valuable so on the whole I would say the benefits outweigh the the cost of the extra storage.

EXT_BeamPhysics.md
DavidSagan: This is a typo. Should it should be a string.

EXT_BeamPhysics.md
DavidSagan: Yes!

EXT_BeamPhysics.md
DavidSagan: Actually this should be moved to the `Non Per-Particle Records of the Particle Root Group` and become a record.

EXT_BeamPhysics.md
DavidSagan: Actually this should be moved to the `Non Per-Particle Records of the Particle Root Group` and become a record.

And the same for `totalCharge`.

EXT_BeamPhysics.md
DavidSagan: Yes.

EXT_BeamPhysics.md
DavidSagan: @ax3l @RemiLehe 

I see I was not consistent about this. I thought that using a trailing slash was the convention for records. It does not matter to me.  Whatever the convention is should be used.

EXT_BeamPhysics.md
DavidSagan: For photons the signal that a detector pixel sees is a sum of the fields of the photons falling upon it. I'm not sure how you would use this parameter in the case of something that was not a photon.

EXT_BeamPhysics.md
DavidSagan: This is indeed similar to `weighting`. There is also in ED-PIC `weightingPower` which gets multiplied with `weighting` and there is no equivalent to `weightingPower` for Beam Physics. All-in-all I would say that trying to combine these in a third extension is more problematical than it is worth. However, renaming `macroCharge` to `weighting` and making the the definition here consistent with ED-PIC would be OK with me.

EXT_BeamPhysics.md
DavidSagan: See below. **z** is the 5th component of the particle's phase space position and there is no simple relationship between s and z.

EXT_BeamPhysics.md
RemiLehe: Sounds good.

EXT_BeamPhysics.md
RemiLehe: OK, I think that these definitions could be included in the base standard. 
Since this is a bit of a different topic, I suggest removing them from the current pull request, and opening another pull request where these definitions are added in the base standard. Would that be okay with you?

EXT_BeamPhysics.md
RemiLehe: Great! This clarifies the ambiguity.

EXT_BeamPhysics.md
RemiLehe: Sounds good to me.

EXT_BeamPhysics.md
RemiLehe: Ok, sorry I wasn't very clear ; when mentioning "trying to store too much information", the main disadvantage I had in mind was not the storage space (which, as you mentioned, is minimal), but rather the increase in complexity/robustness of the standard. 
In particular, here I am worried about robustness when moving the openPMD file (and its "attached" lattice file) on a different machine. Also: should the lattice file obey a particular standard?

EXT_BeamPhysics.md
RemiLehe: @ax3l Do we actually have any convention on this?

EXT_BeamPhysics.md
RemiLehe: OK, your suggested solution sounds reasonable to me.

RemiLehe:
@DavidSagan Thanks for your answers!
For the points on which we both agreed: could you update your pull request? To do so, you can simply create more commits on the same branch, and push them to this repository ; the pull request will update automatically.

DavidSagan:
@RemiLehe 

> For the points on which we both agreed: could you update your pull request? To do so, you can simply create more commits on the same branch, and push them to this repository ; the pull request will update automatically.

Will Do.

EXT_BeamPhysics.md
DavidSagan: OK I will work on it...

EXT_BeamPhysics.md
ChristopherMayes: @DavidSagan the phrase "relative to" could mean an additive factor, or a multiplicative factor. Where do we define what this means?

EXT_BeamPhysics.md
DavidSagan: This does not impact the robustness of the Standard since data reader programs will not be affected. That is, this does not affect the reading of data in the file. 

There will be cases where program A creates a particle data file which is handed off to program B but program B uses a different lattice input format so program B will be using a different lattice file in any case. This is something accelerator physicists have to deal with all the time.

The point is that storing the information as to the lattice file can be useful information. Just like storing  the creation `date`, `machine` and other parameters as specified by the Standard.

EXT_BeamPhysics.md
DavidSagan: @RemiLehe On second thought I believe we do need the version number since how else do we know what version of the Standard a given version of a file is associated with? Of course the version number here should be 2.0.0. @ax3l: Do you agree?

EXT_BeamPhysics.md
DavidSagan: Now is issue #187

EXT_BeamPhysics.md
DavidSagan: Fixed in new draft to be uploaded soon...

EXT_BeamPhysics.md
DavidSagan: Fixed in new draft to be uploaded soon...

EXT_BeamPhysics.md
DavidSagan: Merged into issue #187

EXT_BeamPhysics.md
DavidSagan: Fixed in new draft to be uploaded soon...

EXT_BeamPhysics.md
DavidSagan: Fixed in new draft to be uploaded soon...

EXT_BeamPhysics.md
DavidSagan: Fixed in new draft to be uploaded soon...

EXT_BeamPhysics.md
DavidSagan: Fixed in new draft to be uploaded soon...

EXT_BeamPhysics.md
DavidSagan: Fixed in new draft to be uploaded soon...

EXT_BeamPhysics.md
DavidSagan: Fixed in new draft to be uploaded soon...

EXT_BeamPhysics.md
DavidSagan: @RemiLehe @ax3l: I checked, the base standard is also not consistent on this.

EXT_BeamPhysics.md
DavidSagan: I am expanding on this. In the new draft there is:

  - Attributes: The following attributes are defined.
    - `relative`:
      - Type: Required **(bool)**
      - Description: Is the energy relative to the `referenceEnergy`?
    - `normalized`:
      - Type: Required **(bool)**
      - Description: Is the energy normalized by the `totalMomentum`?
  - Examples: [E = Energy, E0 = `referenceEnergy`, p0 = `totalMomentum`]
    - `relative` = `F`, `normalized` = `F`  --> Energy = E
    - `relative` = `F`, `normalized` = `T`  --> Energy = E / p0
    - `relative` = `T`, `normalized` = `F`  --> Energy = E - E0
    - `relative` = `T`, `normalized` = `T`  --> Energy = (E - E0) / p0

EXT_BeamPhysics.md
DavidSagan: Changed to `weighting` in the new draft to be uploaded soon...

EXT_BeamPhysics.md
DavidSagan: That should have been `referenceTotalMomentum`. changed in the new draft to be uploaded soon...

EXT_BeamPhysics.md
DavidSagan: This ties in with what the convention should be used. See above

EXT_BeamPhysics.md
DavidSagan: @ChristopherMayes: In the upcoming revision I have:

- Attributes: The following attributes are defined.
      - `normalized`:
        - Type: Required **(bool)**
        - Description: Is the energy normalized by the `referenceTotalMomentum`? That is, is `px` equal to Px or Px/P0 (where Px is the true momentum component and P0 is `referenceTotalMomentum`)?
    - Note: A program using phase space coordinates to describe particle positions will typically use the transverse momentum `px` and `py` along with the `totalMomentum` or `energy` records.

EXT_BeamPhysics.md
DavidSagan: In the upcoming revision this information has been put in the **Definitions** section

DavidSagan:
@RemiLehe @ChristopherMayes @ax3l @marcguetg 

OK I pushed a new version based upon the discussion. Much thanks for the comments.

marcguetg (commented):
I really like the way it is now. One small comment though:

- I would make the presence of either charge/totalCharge 'Required' as this would allow processing of most data even if 'SpeciesType' is not implemented.

DavidSagan:
@marcguetg 

> even if 'SpeciesType' is not implemented

I'm not understanding this. The plan is to have the species type extension become official in the 2.0.0 standard along with the Beam Physics standard.

marcguetg:
@DavidSagan 

I understand, I'm just saying this duplication of information would allow a user to implement a 'light version' which is still according to specs. I agree it would be small use case, as most would use the official bindings, but still something that might be worth considering.

EXT_BeamPhysics.md
ax3l: You know that modern file formats such as `HDF5` support references to groups/data sets in external files?

  https://support.hdfgroup.org/HDF5/doc/TechNotes/ExternalFiles.html

  http://davis.lbl.gov/Manuals/HDF5-1.8.7/UG/09_Groups.html

There is no need to reinvent this in openPMD on a higher level - it's fully [transparent](https://en.wikipedia.org/wiki/Transparency_(human%E2%80%93computer_interaction)) to the user / data reader.

For (inefficient) "file format backends" that do not support such references, the data can just be (inefficiently) included.

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
+Extension to the openPMD Standard for Describing Particle Beams and X-rays
+==========================================
+
+Version 2.0.0
+
+Overview
+------------
```
ax3l: Alignment (sorry - markdown specs...)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+
+The `BeamPhysics` extension to the openPMD standard is meant for describing particles and fields commonly encountered
+in accelerator physics simulations.
+
+How to Use this Extension
+---------------
```
ax3l: Alignment (sorry - markdown specs...)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+
+The `BeamPhysics` extension to the openPMD standard is indicated in a data file by the setting of the `openPMDextension` attribute:
+```
+  openPMDextension = BeamPhysics, SpeciesType
```
ax3l: to be in the correct syntax it would be

openPMDextension = `BeamPhysics;SpeciesType`

but maybe the Pseudocode example is a bit confusing here and it's clear as well in words.

EXT_BeamPhysics.md
DavidSagan: @ax3l: I don't understand how your comment relates to including a lattice file name. Can you elaborate?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+- **Global coordinate system**: The **global** coordinate system is a coordinate system that is
+used to describe the position and orientation of machine elements in space. That is, the **global**
+coordinate system is fixed with respect to the building or room where the machine is placed independent of the
+machine itself.
+
+- **Lattice coordinate system**: The curvilinear coordinate system whose "longitudinal"
+coordinate (**s**) typically runs through the nominal centers of the elements
+in the machine. The **lattice** coordinate system is often used to describe particle positions.
+
+- **Macro-particle**: Macro-particles are simulation particles that represent multiple real particles.
```
ax3l: Imho, general macro-particles are not more than *interacting test-particles* or more general *markers* of the Boltzmann-distribution with a weighting.

They do not (need to) represent (or behave like) real particles for their sampling character and if they can be used as representative tracks depends heavily on the methods & integrators. But that's an off-topic discussion ;-)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+  - Description: The location of the root lattice file.
+
+
+`Particle Root Group` Attributes
+---------------------
+
+For each **particle root group** the following attributes are defined:
+
+- `SpeciesType`
+  - Type: Required *(string)*
```
ax3l: yes, that's doubled. maybe we skip the `type` and only add a `Description:` with the reference to the other extension?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+  - Type: Optional *(string)*
+  - Description: The location of the root lattice file.
+
+
+`Particle Root Group` Attributes
+---------------------
+
+For each **particle root group** the following attributes are defined:
+
+- `SpeciesType`
```
ax3l: lower case `s` :-)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+
+- `charge`
+  - Type: Optional *(int)*
+  - Description: The charge state of the particles. Not needed if the charge can be computed
+  from knowledge of the `SpeciesType`.
+
+- `latticeElementName`
+  - Type: Optional *(string)*
+  - Description: The name of the lattice element the particle or particles are in. This only makes sense if all
+  particles are in the same lattice element. Also see: `latticeElementID` and `locaitonInElement`.
```
ax3l: typo: location

EXT_BeamPhysics.md
DavidSagan: Yes

EXT_BeamPhysics.md
DavidSagan: @ax3l : Sorry I don't understand this.

EXT_BeamPhysics.md
DavidSagan: Changed to use semi-colon.

EXT_BeamPhysics.md
DavidSagan: Will be corrected in the next iteration...

EXT_BeamPhysics.md
DavidSagan: Correct. Will be fixed in the next iteration...

EXT_BeamPhysics.md
DavidSagan: It needs to be at least stated that this is required.

DavidSagan:
New revision with minor typo fixes. One thing that needs to be done is reorganization of the Definitions section in conjunction with implementing Definitions section in the base standard #187.

DavidSagan:
Sorry hit the close button by mistake.

EXT_BeamPhysics.md
RemiLehe: I think that @ax3l meant that you should use
```
Overview
--------
```
instead of
```
Overview
------------
```
(i.e. The number of dashes should equal the number of characters in "Overview")

DavidSagan: @ax3l Ah! Now I understand. Thanks.

EXT_BeamPhysics.md
ax3l: yes, this is generally mentioned above in the `Note: ...` of `openPMDextension`

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
+Extension to the openPMD Standard for Describing Particle Beams and X-rays
+==========================================
```
ax3l: Monospace alignment here as well :)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+    - Type: Optional *(string)*
+    - Description: Normalization used for momentum values.
+    - Possible values:
+        - `referenceTotalMomentum`: Normalize with respect to the `referenceTotalMomentum` attribute
+        - `referenceEnergy`: Normalize with respect to the `referenceEnergy` attribute.
+        - `none`: No normalization.
+
+
+Per-Particle Records of the `Particle Root Group`
+---------------------
```
ax3l: markdown does require us to use the same monospace alignment as the section title length

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+      - `relative` = `T`, `normalized` = `T`  --> `totalMomentum` = (P - P0) / p0
+
+
+- `weighting/`
+    - Type: Optional *(float)*
+    - Description: If macro-particles are being simulated, the `weighting` is the the total charge or total number of the collection of particles that a macro-particle represents.
+
+
+Non Per-Particle Records of the `Particle Root Group`
+---------------------
```
ax3l: see above: alignment :)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+- `referenceTotalMomentum`
+  - Type: Optional *(float)* attribute
+  - Description: Specifies the reference total momentum from which the total momentum is measured with respect to.
+
+- `totalCharge`
+  - Type: Optional *(float)* attribute.
+  - Description: The total charge of all the particles.
+
+Particle Record Dataset Attributes
+-----------------------------
```
ax3l: as above: alignment :)

ax3l (commented):
Added a few more inline notes on consistent style and markdown.

Just for consistency with existing extensions (e.g. `ED-PIC` please keep spellings similar. Examples:
- lowercase `type:`, `description:`, `example:` ... in keywords
- `scope:` for required/optional

EXT_BeamPhysics.md
DavidSagan: @ax3l Actually this page https://daringfireball.net/projects/markdown/syntax#header indicates that any number of dashes will work.

DavidSagan:
Note: There is a conflict with `position` and `positionOffset` in the base standard (#182) which must be rectified.

EXT_BeamPhysics.md
ax3l: oh, interesting. Yes, [rst](http://www.sphinx-doc.org/en/stable/rest.html) (reStructuredText) is the one that is strict.

Sorry for that! In case you don't mind: for aesthetic reasons it still looks better to align it ;-)

DavidSagan: Fine with me.

ax3l:
@DavidSagan but you only use `position` and not `positionOffset` in this extension. Anyway, we can discuss this in an hour in detail :)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+    - Type: Optional *(float)*
+    - Description: Length that a particle has traveled.
+
+- `position`
+    - Type: Required 3-vector *(float)*
+    - Description: Position relative to the coordinate origin.
+    - Components: (`x`, `y`, `<z-coord>`) where `<z-coord>` is one of:
+        - `z`: True longitudinal coordinate.
+        - `-beta.c.dt`: Phase space coordinate conjugate to a momentum based "pz".
+        - `-beta.c.t`:Phase space coordinate conjugate to a momentum based "pz".
```
ax3l: missing space after `:`

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+- `latticeElementID`
+  - Type Optional *(string)*
+  - Description: The ID string for the lattice element given by `latticeElementName`. The idea is that while more than
+    one lattice element may have the same name, the ID string will be unique.
+  - This, along with `locationInElement` sets the origin for specifying particle positions in the **lattice** coordinate system.
+  - Example: With [Bmad](https://www.classe.cornell.edu/bmad/) based programs the ID string is of the form
+    **branch-index>>element-index** where **branch-index** is the associated
+branch index integer, and **element-index** is the associated lattice element index within the branch.
+
+- `locationInElement`
```
RemiLehe: Should this be a particle record (one value per particle)?

RemiLehe: Also, this could be encoded as an integer + a float

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,278 @@
...
+  - The origin is always on the longitudinal axis (x = y = 0) of the **lattice** coordinate system.
+  - Possible values:    
+    - `Upstream-End`: Upstream end of element outside of any edge fields.
+    - `<s-position>`: Where `<s-position>` is a number. Inside the element at a distance, given by `<s-position>`,
+    from the upsteam end of the element.
+    - `Downstream-End`: Downstream end of the element outside of any edge fields.
+  - Note: Since some programs will model edge fields of a lattice element as having zero length, the longitudinal **s**-position
+does not necessarily provide enough information to determine where a particle is with respect to an edge field so `Upstream-End` and `Downstream-End` are provided.
+
+- `momentumNormalization`
```
RemiLehe: Is this a duplication? It seems that `momentum` contains attributes such as `normalised`, etc.

EXT_BeamPhysics.md
DavidSagan: OK changed in the coming revision. Note: Markdown does not *require* this.

EXT_BeamPhysics.md
DavidSagan: Modified in the coming revision.

EXT_BeamPhysics.md
DavidSagan: @ax3l: In the coming revision I have changed this to:

> **Macro-particle**: Macro-particles are simulation particles that represent multiple real particles. The number of real particles that a macro-particle represents affects the calculation of the field due to a macro-particle but does not affect tracking.

If you can think of a better name than **macro-particle** please suggest it.

DavidSagan:
@ax3l: @ChristopherMayes @RemiLehe @marcguetg @jlvay 

>Just for consistency with existing extensions (e.g. ED-PIC please keep spellings similar. Examples:
> - lowercase type:, description:, example: ... in keywords
> - scope: for required/optional

Actually I would propose to convert the base standard and extensions to use capitalizations (beginnings of a sentence should be capitalized), and combine `Scope` with `Type` (which makes the text more compact and, I believe, more readable).

For example see https://tools.ietf.org/html/rfc5226 .

EXT_BeamPhysics.md
DavidSagan: Changed in next revision.

EXT_BeamPhysics.md
DavidSagan: Changed in next revision.

EXT_BeamPhysics.md
DavidSagan: Changed in next revision.

EXT_BeamPhysics.md
DavidSagan: Fixed in next revision.

DavidSagan:
@ax3l: @ChristopherMayes @RemiLehe @marcguetg @jlvay

Right now I am defining a `referenceTotalMomentum` where `Total` in this case is meant to emphasize the momentum here is a magnitude and not a vector. I am thinking this might be a bit confusing and whether  `referenceMomentum` would be better. Thoughts?

EXT_BeamPhysics.md
DavidSagan: @RemiLehe: OK take a look at how I reorganized this.

DavidSagan:
@ax3l: @ChristopherMayes @RemiLehe @marcguetg @jlvay

A new version has been pushed.

I have reorganized things based upon everyone's comments and some problems I found with the logic. Thanks everyone for your help! Please take a look.

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,264 @@
...
+
+How to Use this Extension
+-------------------------
+
+The `BeamPhysics` extension to the openPMD standard is indicated in a data file by the setting of the `openPMDextension` attribute:
+```
+  openPMDextension = BeamPhysics;SpeciesType
+```
+
+Note: The `speciesType` extension must be used when using the `BeamPhysics` extension.
```
ax3l: little typo: missing capital `S` in `speciesType` extension

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,264 @@
...
+  - Description: Reference momentum Possibly used for normalizing particle momentum values.
+
+- `referenceTotalEnergy`
+  - Type: Optional *(string)*
+  - Description: Reference total (kinetic + rest mass) energy. Possibly used for normalizing particle momentum values.
+
+  - `speciesType`
+    - Type: Required *(string)*
+    - Description: The name of the particle species. Species names must conform to the
+    `speciesType` extension.
```
ax3l: little typo: missing capital `S` in `speciesType` extension

(the attribute name is correct, which starts lower-case)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,264 @@
...
+  - Possible values:    
+    - -1: Upstream end of element outside of the upstream fringe edge.
+    - 0: Inside the element.
+    - 1: Downstream end of the element outside outside the downstream fringe edge.
+
+- `momentum/`
+  - Type: Optional 3-vector *(float)*
+  - Description: The momentum vector of the particles.
+  - Components: (`px`, `py`, `pz`).
```
RemiLehe: This is slightly inconsistent with the `ED-PIC` extension (https://github.com/openPMD/openPMD-standard/blob/latest/EXT_ED-PIC.md), which uses `momentum/x`, `momentum/y`, `momentum/z`. 

While the extension are not strictly required to be consistent, in practice it would be great if they are. @DavidSagan Would it be fine to change this to `x` ,`y`, `z`?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,264 @@
...
+  - Description: The program generating the data file may model a lattice element using a "hard edge" model where the fringe fields at the ends of the element are modeled as having zero longitudinal length. In such a case, if a particle is at the end of the lattice element, it is important to know if the particle is outside of the fringe or if the particle is inside the fringe within the body of the element. Note that with a hard edge fringe, the longitudinal **s**-position does not necessarily provide enough information to determine where a particle is with respect to an edge field. Another situation where `locationInElement` is useful is with zero length elements that affect the particle transport (such as zero length multipole elements). If the program generating the data file does **not** use any hard edge models or zero length non-marker elements, `locationInElement` should not be present since this parameter is meaningless in this case.
+  - Possible values:    
+    - -1: Upstream end of element outside of the upstream fringe edge.
+    - 0: Inside the element.
+    - 1: Downstream end of the element outside outside the downstream fringe edge.
```
RemiLehe: There is a duplication of `outside` in the above text.

RemiLehe (changes requested):

EXT_BeamPhysics.md
ax3l: VC discussion:
- maybe name `momentum` a `canonicalMomentum`?
- probably ok to be different to `ED-PIC` since they won't fit well together right now in one openPMD series anyway

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,264 @@
...
+  - Type: Optional *(string)*
+  - Description: The location of the root lattice file.
+
+
+`Particle Root Group` Attributes
+--------------------------------
+
+For each **particle root group** the following attributes are defined:
+
+- `charge`
```
RemiLehe: Should this be a per-particle record?

RemiLehe (dismissed):

EXT_BeamPhysics.md
ax3l: VC: let us make this a *scalar record* and use a *constant* scalar record if it's the same for all particles of a species.

EXT_BeamPhysics.md
ax3l: VC: does not apply here, the file is just an ASCII file

EXT_BeamPhysics.md
DavidSagan: Actually this may not be the canonical momentum. For example, in a finite magnetic field there is a term due to the finite vector potential. Which brings up the question as to what to do if someone wants to actually store the canonical momentum when there is a finite magnetic field. In this case the extension will need to be revised to add room for the vector and scalar potentials which can be normalized or unnormalized.

EXT_BeamPhysics.md
DavidSagan: Since `speciesType` is not per-particle I did not make `charge` per-particle. But it does not hurt so I will move it in the next update.

DavidSagan: OK.

DavidSagan:
@RemiLehe @ax3l 👍 
The latest revision is a small change based upon your comments.
-- Cheers, David

EXT_BeamPhysics.md
DavidSagan: Fixed

EXT_BeamPhysics.md
DavidSagan: Fixed

EXT_BeamPhysics.md
DavidSagan: Fixed

EXT_BeamPhysics.md
```diff
@@ -161,12 +161,12 @@ The following records store data on a particle-by-particle basis.
...
           - `z`: `z` component is the true longitudinal coordinate.
 
-- `refTime/`
+- `refTime`
     - Type: Optional *(float)*
     - Description: The reference particle time.
     Note that the reference time is a function of **s** and therefore can be different for different particles.
 
-- `s-Position/`
+- `s-Position`
```
ax3l: Just a style question: really `-` and camelCase in this name?
(similar in `time-refTime`)

DavidSagan: I was not thinking. To be consistent that should be `sPosition`.  I will make the change. The `time-refTime` is to emphasize that this is a time difference. Do you have an alternative name?

ax3l:
@RemiLehe let us take a review on the latest changes :)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,252 @@
+Extension to the openPMD Standard for Describing Particle Beams and X-rays
+==========================================================================
+
+Version 2.0.0
```
ax3l: Let us not write the version here again, since the standard otherwise needs to be updated at several positions at once when bumping it.

Instead, write prominently the extension name:
```suggestion
openPMD extension name: `BeamPhysics`
```

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,252 @@
...
+
+Overview
+--------
+
+The `BeamPhysics` extension to the openPMD standard is meant for describing particles and fields commonly encountered
```
ax3l: feel free to keep latex-like one sentence per line (preferred).
(Alternatively, please break all at 80 char.)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,252 @@
...
+
+- **Phase Space Coordinates**: Phase space coordinates are ordered (x, px, y, py, z, pz).
+
+Notes:
+------
+
+- When using the **lattice** coordinate system, the `position` coordinates are **(x, y, s)** or **(x, y, z)** where, nominally, **x** is the "horizontal" component, **y** is the "vertical" coordinate, and **s** or **z** is the lattice longitudinal coordinate.
+
+Additional File Root Group (path `/`) records
+------------------------------------------------
```
ax3l, line 47: oversized `---` (pls align)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,252 @@
...
+- **Particle Root Group**: The **Particle Root Group** is a group for specifying a group of particles. See the Base Standard for more information.
+
+- **Phase Space Coordinates**: Phase space coordinates are ordered (x, px, y, py, z, pz).
+
+Notes:
+------
+
+- When using the **lattice** coordinate system, the `position` coordinates are **(x, y, s)** or **(x, y, z)** where, nominally, **x** is the "horizontal" component, **y** is the "vertical" coordinate, and **s** or **z** is the lattice longitudinal coordinate.
+
+Additional File Root Group (path `/`) records
```
ax3l: Title spelling:
```suggestion
Additional File Root Group (Path `/`) Records
```

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,252 @@
...
+- `latticeName`
+  - Type: Optional *(string)*
+  - Description: The name of the lattice.
+
+- `latticeFile`
+  - Type: Optional *(string)*
+  - Description: The location of the root lattice file.
+
+
+`Particle Root Group` records
```
ax3l: Title:
```suggestion
`Particle Root Group` Records
```

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,252 @@
...
+--------------------------------
```
ax3l: oversized `---`... (pls align)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,252 @@
...
+    - Description: The reference particle time.
+    Note that the reference time is a function of **s** and therefore can be different for different particles.
+
+- `sPosition`
+    - Type: Optional *(float)*
+    - Description: Longitudinal distance of the particle in **lattice** coordinates. This record establishes
+    the origin for the (x, y) transverse plane where the particle is measured with respect to in the **lattice** coordinate system.   
+    - Attribute:
+      - `absolute`:
+        - Type: Required **(bool)**
```
ax3l:
```suggestion
        - Type: Required *(bool)*
```

ax3l (commented):
Little style fixes.

Can you by coincidence share an example file for the extension to look at?

EXT_BeamPhysics.md
DavidSagan: @RemiLehe Done.

EXT_BeamPhysics.md
DavidSagan: @ax3l You lost me what does "VC" mean?

EXT_BeamPhysics.md
DavidSagan: Chris Mayes came up with a better way of doing things here so this thread is moot.

EXT_BeamPhysics.md
DavidSagan: Fixed in next iteration coming soon...

EXT_BeamPhysics.md
DavidSagan: Fixed in next iteration coming soon...

EXT_BeamPhysics.md
DavidSagan: Fixed in next iteration coming soon...

EXT_BeamPhysics.md
DavidSagan: Fixed in next iteration coming soon...

EXT_BeamPhysics.md
DavidSagan: @ax3l 
At this point in time we are talking about having to modify 4 files each time the version is changed. The other side of the argument is that if someone stores a standards file outside of GIT then later on it will be easy to see which version the file is for. This is not a big deal for me so if you do think it is better then I will go along and remove it. What do you think?

EXT_BeamPhysics.md
DavidSagan: @ax3l 
Right now mostly long paragraphs are a single line. I don't mind to break at 80 characters but I do not know how to do convert the file efficiently.

EXT_BeamPhysics.md
ax3l: Yes, let's keep it only in the main file for now :)

EXT_BeamPhysics.md
ax3l: sorry, I just noted what we discussed offline in our video conference (VC)

EXT_BeamPhysics.md
ax3l: That's fine for me, just use what is typically understood by people of the domain the extension is written for.

EXT_BeamPhysics.md
ax3l: I mean: just keep one sentence per line and break after each sentence.
This will keep diffs smaller and is the least stressful while writing and editing.

EXT_BeamPhysics.md
ax3l: What we discussed was: should we include the "lattice file", which can be constant for a whole simulation, in each file. I mentioned that HDF5 supports "external files" for this purpose, just like links, which avoid writing the same data over and over again. But then we realized that the lattice file is not an openPMD or HDF5 file but a plain ASCII file.
The question that remains: should we specify how a "lattice file" looks here? It's probably not clear to a reader (at least it is not to me) :)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,348 @@
...
+  - Description: Magnetic field components using rectangular coordinates. Used with `gridGeometry` set to `rectangular`. If the field is DC (`harmonic` is zero), only the real component should be nonzero.
+
+
+- `Ex`, `Ey`, `Ez`
+  - Type: Optional (Either all must be present or all must be absent) *(complex)*
+  - Description: Electric field components. Used with `gridGeometry` set to `xyz`.If the field is DC, only the real component should be nonzero.
+
+
+Per-grid `External Fields Group` Records for `(r, theta, z)` Grids
+--------------------------------------------------------------
```
ax3l: just a tiny alignment issue with markdown here.

EXT_BeamPhysics.md
DavidSagan: Done

EXT_BeamPhysics.md
DavidSagan: OK

EXT_BeamPhysics.md
ax3l: Just updated in another thread we discussed: we keep it in the main document for now.

DavidSagan: Axel asked for removal so the version has been removed.

EXT_BeamPhysics.md
ax3l: Oh, I am fine with unifying this. I probably try to use a trailing `/` so far when I am referring to non-scalar records.

EXT_BeamPhysics.md
DavidSagan: Done

EXT_BeamPhysics.md
ax3l: Maybe `refTimeDifference`?

EXT_BeamPhysics.md
ax3l: somehow this is still two lines after the update :)

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+- When using the **lattice** coordinate system, the `position` coordinates are **(x, y, s)** or **(x, y, z)** where, nominally, **x** is the "horizontal" component, **y** is the "vertical" coordinate, and **s** or **z** is the lattice longitudinal coordinate.
+
+Additional File Root Group (Path `/`) Attributes
+------------------------------------------------
+
+The following records are defined for the file root group.
+
+- `fileType`
+  - Type: Optional *(string)*
+  - Description: The type of data being stored in the file. If present, must be set to `openPMD`. This attribute is used in systems where different data files can contain different types of data and allows for quick identification of the what type of data is in a given file.
```
ax3l: maybe I missed it, but isn't this already possible to be quickly identified with the base standard identifier, attribute `openPMD` in `/`?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+- `latticeName`
+  - Type: Optional *(string)*
+  - Description: The name of the lattice.
+
+- `latticeFile`
+  - Type: Optional *(string)*
+  - Description: The location of the root lattice file.
+
+Particle Group Standard
+============================
```
ax3l:
```suggestion
=======================
```
cosmetic

EXT_BeamPhysics.md
DavidSagan: >The question that remains: should we specify how a "lattice file" looks here? It's probably not clear to a reader (at least it is not to me) :)

Standardization of the lattice file is not possible. For example, the syntax of a lattice file for use by ASTRA is different from Bmad.

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+
+The **Particle Group** is a group for specifying a set of particles such as the particles in a bunch. Multiple **Particle Groups** can be defined in a file. The path for a **Particle Group** is given by the **basePath** and **particlesPath** attributes in the file root group as discussed in the base OpenPMD standard. For example, if **basePath** is  "/data/%T/", and **ParticlesPath** is "particles/", then **Particle Groups** paths would be of the form "/data/%T/particles/" where "%T" is an integer. EG: "/data/37/particles/".
```
ax3l:
```suggestion
The **Particle Group** is a group for specifying a set of particles such as the particles in a bunch. Multiple **Particle Groups** can be defined in a file. The path for a **Particle Group** is given by the **basePath** and **particlesPath** attributes in the file root group as discussed in the base OpenPMD standard. For example, if **basePath** is  `/data/%T/`, and **ParticlesPath** is `particles/`, then **Particle Groups** paths would be of the form `/data/%T/particles/` where `%T` is an integer. EG: `/data/37/particles/`.
```
cosmetic

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+
+
+`Particle Group` Attributes
+--------------------------------
```
ax3l, line 70:
```suggestion
---------------------------
```
cosmetic

EXT_BeamPhysics.md
DavidSagan: The extension now uses trailing slash everywhere.

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+
+- `chargeUnitSI`
+  - Type: *(real)*. Only required if `chargeLive` or `totalCharge` is present.
+  - Description: Unit conversion factor to multiply `chargeLive` or `totalCharge` by in order to convert to SI units.
+
+- `latticeElementName`
+  - Type: Optional *(string)*
+  - Description: The name of the lattice element the particle are in. This only makes sense if all particles are in the same element. [Keep in mind that if particles are lost and the lost particles are also included in the file, not all particles may be in the same element.] Also see: `locationInElement`.
+
+- `numParticles`
```
ax3l, line 86: Is this strictly needed or can this be inferred by just querying any record in the particle group and its file-format provided extent?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+  - Type: Required *(string)*
+  - Description: The name of the particle species. Species names must conform to the `SpeciesType` extension.
+  - Example: `electron`, `H2O`.
+
+- `totalCharge`
+  - Type: Optional *(real)*
+  - Description: The total charge of all the particles alive and dead.
+
+Per-Particle Records of the `Particle Group`
+-------------------------------------------------
```
ax3l:
```suggestion
--------------------------------------------
```
cosmetic

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+    - Components: (`x`, `y`, `z`).
+
+- `locationInElement`
+   - Type Optional *(integer)*
+   - Description: The program generating the data file may model a lattice element using a "hard edge" model where the fringe fields at the ends of the element are modeled as having zero longitudinal length. In such a case, if a particle is at the end of the lattice element, it is important to know if the particle is outside of the fringe or if the particle is inside the fringe within the body of the element. Note that with a hard edge fringe, the longitudinal **s**-position does not necessarily provide enough information to determine where a particle is with respect to an edge field. Another situation where `locationInElement` is useful is with zero length elements that affect the particle transport (such as zero length multipole elements). If the program generating the data file does **not** use any hard edge models or zero length non-marker elements, `locationInElement` should not be present since this parameter is meaningless in this case.
+   - Possible values:    
+     - -1: Upstream end of element outside of the upstream fringe edge.
```
ax3l:
```suggestion
     - `-1`: Upstream end of element outside of the upstream fringe edge.
```
cosmetic

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+     - 0: Inside the element.
```
ax3l:
```suggestion
     - `0`: Inside the element.
```
cosmetic

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+     - 1: Downstream end of the element outside the downstream fringe edge.
```
ax3l:
```suggestion
     - `1`: Downstream end of the element outside the downstream fringe edge.
```
cosmetic

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+- `totalMomentumOffset/`
+  - Type: Optional *(real)*
+  - Description: Base total momentum from which `totalMomentum` is measured. That is, True total momentum = `totalMomentum + totalMomentumOffset`. Some programs will use `totalMomentumOffset/` to store the **reference momentum** in which case `totalMomentum` will then be the deviation from the referece.
+
+- `particleCoordinatesToGlobalTransformation/`
+  - Type: Optional group.
+  - Description: Defines the transformation from the coordinates used to describe a particle to the **global** coordinate system.
+  - `R_frame`:
+    - Required 3-vector *(real)* Attribute
+    - Description: specifying the (x, y, z) position of the coordinate origin that the particles are measured with respect to in the **global** coordinate frame.
```
ax3l:
```suggestion
    - Description: specifying the (`x`, `y`, `z`) position of the coordinate origin that the particles are measured with respect to in the **global** coordinate frame.
```
cosmetic for consistency

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+  - Description: Base path to the external mesh field groups. Use the **%T** construct if there are multiple meshes present.
+  - Example: `/ExternalFieldMesh/%T/`. Base paths to the external fields group, in this case, would be `/ExternalFieldMesh/1/`, etc.
+  - Example: `/ExternalFieldMesh/`. In this case there is only one external fields group.
+
+Notes
+-----
+
+- AC fields can be described using complex numbers. The actual field is the real part of
+
+    Z &ast; Exp[-2 &ast; pi &ast; I &ast; (f &ast; t + phi)]
```
ax3l: verbatim with triple `'s probably works well, too:
```
    Z * Exp[-2 * pi * I * (f * t + phi)]
```

EXT_BeamPhysics.md
DavidSagan: momentumNormalization no longer exists

EXT_BeamPhysics.md
DavidSagan: But a non-openPMD file will not have the `openPMD` attribute.

EXT_BeamPhysics.md
DavidSagan: Will be in next push...

EXT_BeamPhysics.md
DavidSagan: Will be in next push...

EXT_BeamPhysics.md
DavidSagan: Will be in next push...

EXT_BeamPhysics.md
DavidSagan: Will be in next push...

EXT_BeamPhysics.md
DavidSagan: Now `sPosition`

EXT_BeamPhysics.md
DavidSagan: Unfortunately it cannot always be inferred.

EXT_BeamPhysics.md
DavidSagan: Will be in next push...

EXT_BeamPhysics.md
DavidSagan: Will be in next push...

EXT_BeamPhysics.md
DavidSagan: Will be in the next push...

EXT_BeamPhysics.md
ax3l: I am confused how this is useful for files that use openPMD and the openPMD extension :)

EXT_BeamPhysics.md
DavidSagan: Done.

EXT_BeamPhysics.md
DavidSagan: Done.

EXT_BeamPhysics.md
DavidSagan: Done.

EXT_BeamPhysics.md
DavidSagan: The proposed standard now does not have canonical momentum so this discussion is moot.

EXT_BeamPhysics.md
DavidSagan: It is useful for readers that can read a variety of data types including openPMD.

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,344 @@
+BeamPhysics Extension to the openPMD Standard for Describing Particle Beams, Photons, and External Fields
+=========================================================================================================
+
```
ax3l, line 4: As in `ED-PIC` and `SpeciesType`: can we please make
```md
openPMD extension name: `BeamPhysics`
```

The first line here? Just to keep extensions consistent.
```suggestion

openPMD extension name: `BeamPhysics`
```

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,344 @@
...
+Overview
+--------
+
+The `BeamPhysics` extension to the openPMD standard is meant for describing particles and fields commonly encountered in accelerator physics simulations.
+
+How to Use this Extension
```
ax3l, line 10: We potentially do not need this paragraph on technical usage ("how specify the attribute") since the base standard explains this already.

Instead, we can write an introduction here and the requirement to co-use the extension with `SpeciesType`.

ax3l (commented):
This already looks very good to me. :rocket: :sparkles: 

@DavidSagan please just address the introduction section comments to unify the style with the other extension.
Then, please [`squash` your commits](https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/) into one:
```bash
git fetch --all
git rebase -i origin/upcoming-2.0.0
# change all but the first commit to "squash" instead of "pick"
git push -f
```

@ChristopherMayes and @RemiLehe can you give this your blessing already? :) Comments, wishes?

EXT_BeamPhysics.md
DavidSagan: Will be in next version

EXT_BeamPhysics.md
DavidSagan: @ax3l I'm not exactly sure what you are thinking about here. Could you write a proposed change?

ax3l: No problem, can you squash all commits first please?

EXT_BeamPhysics.md
```diff
@@ -0,0 +1,346 @@
...
+Notes
+-----
+
+- AC fields can be described using complex numbers. The actual field is the real part of
+
+    Z &ast; Exp[-2 pi i f &ast; (t - t0)]
+
+where `Z` is the complex field, `f` is the Oscillation frequency, `t` is the time, and `t0` is a reference time.
+
+Software that writes complex data must use native complex storage for formats that support complex numbers. For formats that do not have native complex support (for example, HDF5), complex numbers are to be stored using subgroups labeled `r` and `i` for real and imaginary respectively.
```
ax3l: I would change this to:
```suggestion
Software that writes complex data must use native complex storage for formats that support complex numbers.
For formats that do not have native complex support (for example, HDF5), complex numbers are to be stored using value pairs for real and imaginary component and shall follow the `h5py` convention for annotations.
```
Example code:
```py
import h5py as h5                                                                                
import numpy as np                                                                               

x = np.array([1.+2j, 3.+4j, 5.+6j], np.cdouble)                                                  
x                                                                                                
# array([1.+2.j, 3.+4.j, 5.+6.j])
x.dtype                                                                                          
# dtype('complex128')
f = h5.File('complex.h5', 'w')                                                                   
f['/mydata'] = x                                                                                 
f.close()
```

Creates:
```bash
$ h5ls -d -r complex.h5 
/                        Group
/mydata                  Dataset {3}
    Data:
        (0) {1, 2}, {3, 4}, {5, 6}

$ h5dump complex.h5
HDF5 "complex.h5" {
GROUP "/" {
   DATASET "mydata" {
      DATATYPE  H5T_COMPOUND {
         H5T_IEEE_F64LE "r";
         H5T_IEEE_F64LE "i";
      }
      DATASPACE  SIMPLE { ( 3 ) / ( 3 ) }
      DATA {
      (0): {
            1,
            2
         },
      (1): {
            3,
            4
         },
      (2): {
            5,
            6
         }
      }
   }
}
}
```

I think this totally works well and corresponds to what I suggest here: https://github.com/openPMD/openPMD-standard/issues/29#issuecomment-566781013

ax3l:
Before we merge this:
Can you please provide an example file or file series that used this convention already? I would like to run it through the validator, viewer, and reference API implementations to see if the extension is fully compatible with the base standard.

EXT_BeamPhysics.md
```diff
@@ -142,6 +142,10 @@ The following records store data on a particle-by-particle basis.
   - description: Base momentum from which `momentum` is measured. That is, True momentum = `momentum + momentumOffset`. Assumed zero if not present.
   - components: (`x`, `y`, `z`).
 
+-`particleID`
```
ax3l: a globally unique particle `id` is already available in the base standard to hook into. Would this work for you?
https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/STANDARD.md#records-for-each-particle-species

DavidSagan: @ChristopherMayes  Does the base standard "id" work for you? Notice that the description is

"description: a globally-unique identifying integer for each particle, that can be used to, e.g., track particles. This identifying integer should be truly unique within the simulation; in particular, even among different particle species, two particles should not have the same id. Also, when a particle exits the simulation box, its identifying integer should not be reassigned to a new particle."

@ax3l  The base standard needs to be modified to not use “Particle Species”. The beam physics extension uses the name “Particle Group”. Also the standard should not be mandating what type of integer should be used as it is for "id".

DavidSagan: @ax3l I should note that my code does not use ID numbers so I am fine with what everyone else wants.

ChristopherMayes: @DavidSagan `id` is fine for me.

DavidSagan: @ax3l Take a look at the change and tell me if this is OK.

ax3l:
offline/zoom discussed: we don't need to squash and rebase, I will just `squash & merge` all PRs from now on :)

EXT_BeamPhysics.md
ax3l: Looks great, good idea to mention the reference again here.

ax3l (approved):
The extension looks like it is in great shape. Thank you for the work on this and the contribution!

@ChristopherMayes @RemiLehe, if you confirm, I would target a merge today. Otherwise if you have further comments early next week. What do you think?

ChristopherMayes (approved):

ax3l:
@marcguetg just as a notification to you: we plan to merge the beam-physics extension

Issue #151, opened 2017-11-22
Why a bitmask of extensions?


t184256:
I have a handful of questions I've asked @ax3l by email some time ago. I've got no answer, so maybe I'll try asking them here.

`openPMDextension` is a measly `uint32` bitmask, why? Why isn't it an set of attributes, an array of strings, or basically anything more flexible/extensible? It may seem to be an non-issue now, but the more it will become one, the harder it'll be to change.

ax3l:
Hi Alexander,

sorry for the late response - I indeed have not forgotten your mail and just did not get to it yet (also because our lab evaluation period just starts in various fields...). I will still try to answer your other questions in the e-mail, but indeed the openPMD issue tracker is the right place to ask: here we can publicly archive, link, check and improve orthogonal questions and improvements :)

`openPMDextension`: openPMD extensions are in principle openPMD extensions could be stacked and used at the same time. E.g. one could write a "plasma" extension that defines that "B" is always naming records that are magnetic fields and on top a PIC and a Hydro extension that define namings for used solvers/methods.

A bitmask is just the most efficient representation of such a possible combination and was taken for the sake of very simple parsing (in C/C++/Fortran). With the property we used they are "extensible" ~~enough~~. Indeed, we could also use a list of (variable length) strings. (E.g. in some places such as the ED-PIC extension we alsways use comma-separated lists.)

## The Proposed Change to a String

This would improve human readability, make writing not much more complicated, introduces a bit of negligible overhead during meta data read and makes parsing a bit more lines in strongly typed languages. Also, named collisions are less likely to occur than number collisions if someone is writing an extension and forgets to ping the issue tracker first for an ID. 🚀 ✨ 👍

ax3l:
CCing @DavidSagan and @CFGrote as they suggest the same change. Thx!

t184256:
I cannot resist the temptation to answer.

> With the property we used they are extensible enough.

A maximum of 32 extensions doesn't sound future-proof.

> A bitmask is just the most efficient representation of such a possible combination and was taken for the sake of very simple parsing

Oh yay, let's save a thousand CPU cycles of overhead and a few dozen of bytes. In a file with megabytes or gigabytes of finely gridded data, where each particle or grid point takes way more than 4 bytes! Sorry, that doesn't sound reasonable.

ax3l:
Ah I should cross-check my writing before submit!
[...]

PR #185, opened 2018-02-20
openPMDextension: Replace ID with String

ax3l:
Replaces the `openPMDextension` attribute with a semicolor-separated list of strings of names.

*Implements issue:*  #151

cc: @t184256 

## Description

This improves readability and does not cost much at runtime. The change both affects the *base standard* and the only extension in `1.1.X`: `ED-PIC`.

Additionally, the attribute is now *optional*, so one only needs to set it if one really wants to use extensions.

## Affected Components

- `base`
- `EXT: ED-PIC`

## Logic Changes

For the pure use of the base standard, this attribute must be skipped.

Instead of a bitmask one now has to write a semicolon-separated list of names in the `openPMDextension` attribute, e.g.:
```python
enable_extensions = ['ED-PIC', 'SpeciesType']

# h5py group
group.attrs["openPMDextension"] = np.string_(';'.join(extensions))
```

## Writer Changes

Writers **must** skip the attribute now if they are not using extensions.

## Reader Changes

Readers **must** check now if the attribute `openPMDextension` exists at all.
If it exists, splitting the resulting string by `;` will lead to a list of names.
These names are the openPMD extensions.

*What would a reader need to change?*

- [x] `openPMD-validator`: https://github.com/openPMD/openPMD-validator/pull/36
- [ ] `openPMD-viewer`: https://github.com/openPMD/openPMD-viewer
- [ ] `yt`: https://github.com/yt-project/yt
- `openPMD-api`: https://github.com/openPMD/openPMD-api/projects/1

## Data Updater

Since `openPMDextension` was required before, all existing files are affected.

- [x] The `openPMD-updater` script already implements the update logic in:

[`openpmd_updater/transforms/v2_0_0/ExtensionString.py`](https://github.com/openPMD/openPMD-updater/blob/9c59cc7af40431f46d5df5955bf7a25a7529e60d/openpmd_updater/transforms/v2_0_0/ExtensionString.py)

### DIFF
--- a/EXT_ED-PIC.md
+++ b/EXT_ED-PIC.md
@@ -3,8 +3,6 @@ Domain-Specific Naming Conventions for Electro-Dynamic/Static PIC Codes
 
 openPMD extension name: `ED-PIC`
 
-openPMD extension ID: `1`
-
 
 Introduction
 ------------
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -77,22 +77,6 @@ Each file's *root* group (path `/`) must at least contain the attributes:
                    minor and revision must not be neglected
     - example: `1.1.0`
 
-  - `openPMDextension`
-    - type: *(uint32)*
-    - description: the bit-mask of unique IDs of applied extensions of the
-                   openPMD standard
-                   (see: *Domain-Specific Extensions*);
-                   to test for a specific extension `ID` perform a bit-wise
-                   `AND` operation on `openPMDextension`:
-                   `extension_1_used = (( openPMDextension AND ID ) == ID)`
-    - note: if only one extension is used at a time, the value for
-            `openPMDextension` is simply the `ID` of the extension
-    - examples:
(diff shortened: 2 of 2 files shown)


DavidSagan (approved):
Looks good.

ax3l:
Btw, I am wondering if we should replace `ED-PIC` with `EDPIC` or `PIC` for consistency?
(I would like to keep the `electro-dynamic` part since *gyro* pic codes are much different.)

RemiLehe (approved):

RemiLehe:
Regarding `ED-PIC`: maybe this should be a separate issue.

Regarding this particular PR: should I merge?

ax3l:
Sure, please go ahead :)

franzpoeschel:
Is there a specific reason why this should be a semicolon-separated list of strings, instead of simply a list of strings? I think that all our backends would support that.

PR #190, opened 2018-03-29
Close #155 Units should not be forced to SI

### LINKED ISSUE #155: Units should not be forced to SI
We [the Smilei developers] would like to propose a change that was discussed a little [somewhere else](https://github.com/SmileiPIC/Smilei/issues/3): **specifying units in SI could be optional**. In particular, this concerns the attributes `timeUnitSI` (in `basePath`), `gridUnitSI` (of each `mesh record`) and `unitSI` (of each `Record Component`).

We know that this possibility has found significant opposition here, but we would like to expose clearly the reasons, and a possible solution.

Why optional units?
----

From our perspective of PIC code developers and users, most of the simulations we run are scalable by some arbitrary factor. This is due to the Maxwell-Vlasov system of equations having a natural set of units: times, lengths, electromagnetic fields, densities and currents may be expressed in a normalized (unit-less) manner. They all depend on each other, but there is no absolute value of these quantities. In fact, the result of the simulation can be scaled using one arbitrary *number*, as long as each of the quantities above have been scaled in the correct way, with the correct exponent. Typically, in the case of the code Smilei, we have chosen this *number* to be in the dimensions of an angular frequency `Wr`. The simulation times are in units of `1 / Wr`, distances in units of `c / Wr`, densities in units of `ε0 m_e Wr^2 / e^2`, etc. The results of the simulations can be scaled by these factors. For instance, if you multiply all times by 2, then you also have to multiply distances by 2, and densities must be divided by 4.

This does not apply to PIC simulations with additional physics which require an absolute system of units (ionization, collisions, QED, etc.) but this case is trivial.

It turns out that many PIC codes disregard the fact that the PIC equations are naturally scalable by some factor, and impose some absolute units. In our opinion this is too restrictive. We don't mean PIC codes should change, but OpenPMD should respect the variety of codes that use meshes and particles in order to broaden its scope.

How to specify the simulation scaling?
----

Electromagnetic PIC codes are not the only ones that can be scaled. It is also the case for Vlasov codes, electrostatic [Poisson-Vlasov] codes, and, to some extend, hydro codes. Even though EM-PIC codes have only one scaling factor, in general, there may be more (e.g. two factors for Poisson-Vlasov codes).

We thus propose new optional attributes as follows:
* in `basePath`, a new `timeUnitScalings` attribute
* in each `mesh record`, a new `gridUnitScalings` attribute
* in each `Record Component`, a new `unitScalings` attribute

These new attributes should each be a list of floats. They must all have the same number of elements. Each element represent the power at which a scaling factor should be taken if a user wants to scale the simulation up or down.

Let's take an example. In a EM-PIC simulation, there is only one scaling factor (in Smilei, the `Wr` above). This means that each of the new attributes above should be a list of one float. In this case, `timeUnitScalings=[-1.]` because times scale as `1/Wr`, `gridUnitScalings=[-1.]` because distances scale as `1/Wr` and for a record component that represents a plasma density, `unitScalings=[2.]` because densities scale as `Wr^2`. In other cases, we may have several scaling factors, so that these lists of scalings may have more than one element.

**New rules**:
* Both `***unitSI` and `***unitScalings` are optional.
* The rule above implies that OpenPMD cannot enforce a comparison between units of data. It is up to the user to make sure they have understood the type of data they are handling.
* The absence of the `***Scalings` attributes means the data is not scalable by any factor. Alternately, it can be decided that an empty list should be defined instead.
* There may be both `***unitSI` and `***unitScalings` defined. The former being seen as the "default" scaling of the simulation, while the latter indicates that the simulation may be freely scaled.
* If any of the new attributes `timeUnitScalings`, `gridUnitScalings`, and `unitScalings` are defined, they must all have the same length.

Note that `Scalings` may not be the best wording for all attributes above. It can be confused with performance scalings (weak/strong scalings) in HPC computing. There could be a better word.

Doesn't this break OpenPMD fundamentals: making data comparable?
----

*If the data of one code cannot be represented in SI units, how are you supposed to compare it to another code?*
This has been one the most frequent arguments against having units optional. Let me explain why we disagree.

If one code provides simulation results in SI, and the other one normalized units, the user has to know what they want to compare and how the data should be modified in order to compare it. This is not the job of OpenPMD, in our opinion.  Many simulations are fundamentally scalable, and later comparison to other simulations must be dealt knowingly by the user. We believe that the new attributes above provide a complete manner to scale the data. OpenPMD clients will have to provide users with a way to introduce scaling factors. In fact, this somewhat ensures that users actually know what they are doing instead of randomly compare things.

Overall, we understand that this is a complex question that changes features in OpenPMD but we believe that it needs to be addressed to make OpenPMD widely used. We hope we have convinced you that the current standard has flaws with respect to units management.

RemiLehe:
## Description

This makes the attributes `unitSI`, `timeUnitSI` and `gridUnitSI` recommended instead of required. This is important for some codes which do not have a well-defined conversion to SI. Closes #155

## Affected Components

- `base`: These attributes are now recommended instead of required.
- `EXT-...`: Any extension that relies on these attributes should now be aware of the fact that it may not exist.

## Writer Changes

Write can optionally omit this attribute (although it is very strongly recommended to add it, unless there is a good reason). 

## Reader Changes
- [x] `openPMD-validator`: Should make these attributes recommended: https://github.com/openPMD/openPMD-validator/pull/37

The readers below should now always check whether these attributes exist. If not, the readers should return the queried quantities without any conversion (i.e. as it is written in the openPMD file). This assumes that the user will understand the unit system. For readers that show the unit symbol (e.g. kg, m, V/m), these standard symbols should not be shown. It is up to each reader to choose how to display the non-converted data.
- [ ] `openPMD-viewer`: Should check before reading; print dimensions on raw data if missing, e.g. `[L * T^-1]`

## Data Converter

No, existing files will have the attributes and do not need to change.

### DIFF
--- a/EXT_ED-PIC.md
+++ b/EXT_ED-PIC.md
@@ -364,7 +364,8 @@ particle. Therefore, this extension requires the two following attributes:
 f = h5py.File('example.h5')
 species = f["<path_to_species_group>"]
 q = species["charge"][:]
-u_si = q.attrs["unitSI"]
+u_si = q.attrs.get("unitSI", default=1.)
+# `default` handles the case where conversion to SI is not provided
 p = q.attrs["weightingPower"]
 if q.attrs["macroWeighted"] == 0 and p != 0:
     w = species["weighting"][:]
--- a/EXT_ED-PIC.md
+++ b/EXT_ED-PIC.md
@@ -364,8 +364,8 @@ particle. Therefore, this extension requires the two following attributes:
 f = h5py.File('example.h5')
 species = f["<path_to_species_group>"]
 q = species["charge"][:]
-u_si = q.attrs.get("unitSI", default=1.)
 # `default` handles the case where conversion to SI is not provided
+u_si = q.attrs.get("unitSI", default=1.)
 p = q.attrs["weightingPower"]
 if q.attrs["macroWeighted"] == 0 and p != 0:
     w = species["weighting"][:]
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -191,6 +191,7 @@ standard:
     - description: an arbitrary comment
     - example: `After each time step we randomly removed 5 particles.`
 
+
 Iterations and Time Series
 --------------------------
 
@@ -507,8 +508,8 @@ def is_const_component(record_component):
 
 def get_component(group, component_name):
     record_component = group[component_name]
-    unitSI = record_component.attrs.get("unitSI", default=1.)
     # `default` handles the case where conversion to SI is not provided
+    unitSI = record_component.attrs.get("unitSI", default=1.)
 
     if is_const_component(record_component):
         return record_component.attrs["value"], unitSI
@@ -590,21 +591,18 @@ patch order:
 Unit Systems and Dimensionality
 -------------------------------
 
-All datasets and attributes can be written in arbitrary units, in the openPMD
-file. Typically, this will be the unit system adopted internally by the
-software that writes the openPMD file (e.g. the unit system adopted internally
+All datasets and attributes can be written in arbitrary units, in openPMD. 
+Typically, this will be the unit system adopted internally by the
+software that writes openPMD (e.g. the unit system adopted internally
 by a physics simulation code), in order to avoid a reduction in performance
 associated with unit conversion during the write process.
 
 However, for each dataset and attribute, it is **very strongly** recommended to
 provide the conversion factor from the chosen unit system to the International
-System of Units (SI). (See the description of the attributes `timeUnitSI`, `gridUnitSI` and `unitSI`, in the present document.) This allows
+System of Units (SI). (See the description of the attributes `timeUnitSI`, 
+`gridUnitSI` and `unitSI`, in the present document.) This allows
 the reading softwares to convert the datasets to units that any user can
-understand, without requiring them to know the chosen units system.
-
-(Note that, for human readable output, it is possible to add the actual string
-of the unit in the corresponding `comment` attribute, for purely informational
-purposes.)
+understand, without requiring them to know the chosen unit system.
 
 In some exceptional cases, the software that produces the openPMD data may
 not have a well-defined conversion to the SI system. (This is the case for
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -95,7 +95,7 @@ Each file's *root* group (path `/`) must at least contain the attributes:
       present standard (e.g. fields on an unstructured mesh),
       this can be done be storing this data within a path that *is not*
       of the form given by `basePath` (e.g. `/extra_data`). In this
-      way, the openPMD parsing tools will not parse this additional data. 
+      way, the openPMD parsing tools will not parse this additional data.
 
 The following attribute is *optional* in each each file's *root* group
 (path `/`) and indicates if a file also follows an openPMD extension
@@ -191,7 +191,6 @@ standard:
     - description: an arbitrary comment
     - example: `After each time step we randomly removed 5 particles.`
 
-
 Iterations and Time Series
 --------------------------
 
@@ -255,6 +254,9 @@ time step.
    - description: The latest time step (that was used to reach this iteration).
                   This is needed at the iteration level, since the time step
                   may vary from iteration to iteration in certain codes.
+
+In addition, the following attribute is *recommended* (see the section
+on Unit Systems and Dimensionality, further below):
 
  - `timeUnitSI`
     - type: *(float64 / REAL8)*
@@ -423,6 +425,9 @@ meshes):
 
     - example: `(0.0, 100.0, 0.0)` or `(0.5, 0.5, 0.5)`
 
+In addition, the following attribute is *recommended* (see the section
+on Unit Systems and Dimensionality, further below):
+
   - `gridUnitSI`
     - type: *(float64 / REAL8)*
     - description: unit-conversion factor to multiply each value in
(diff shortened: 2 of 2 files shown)


RemiLehe:
@beck-llr @mccoys @DavidSagan Let me know me know if you have any comment.

DavidSagan:
@RemiLehe Looks fine with me.

mccoys (approved):

EXT_ED-PIC.md
```diff
@@ -364,7 +364,8 @@ particle. Therefore, this extension requires the two following attributes:
 f = h5py.File('example.h5')
 species = f["<path_to_species_group>"]
 q = species["charge"][:]
-u_si = q.attrs["unitSI"]
+u_si = q.attrs.get("unitSI", default=1.)
+# `default` handles the case where conversion to SI is not provided
```
ax3l: one usually puts a comment before the line :-)

ax3l:
I took the time and implemented the change in the validator in https://github.com/openPMD/openPMD-validator/pull/37

STANDARD.md
```diff
@@ -191,7 +191,6 @@ standard:
     - description: an arbitrary comment
     - example: `After each time step we randomly removed 5 particles.`
 
-
```
ax3l: please keep the double newline, we have it before each headline to structure the document

STANDARD.md
```diff
@@ -95,7 +95,7 @@ Each file's *root* group (path `/`) must at least contain the attributes:
       present standard (e.g. fields on an unstructured mesh),
       this can be done be storing this data within a path that *is not*
       of the form given by `basePath` (e.g. `/extra_data`). In this
-      way, the openPMD parsing tools will not parse this additional data. 
+      way, the openPMD parsing tools will not parse this additional data.
```
ax3l, line 98: although this EOL whitespace is annoying and needs to go, pls avoid changing lines that are not related to the change set so we can keep the chance for collisions with other PRs minimal

STANDARD.md
```diff
@@ -502,7 +507,8 @@ def is_const_component(record_component):
 
 def get_component(group, component_name):
     record_component = group[component_name]
-    unitSI = record_component.attrs["unitSI"]
+    unitSI = record_component.attrs.get("unitSI", default=1.)
+    # `default` handles the case where conversion to SI is not provided
```
ax3l: usually before the line

STANDARD.md
```diff
@@ -584,19 +590,37 @@ patch order:
...
-Units (SI).
+All datasets and attributes can be written in arbitrary units, in the openPMD
+file. Typically, this will be the unit system adopted internally by the
+software that writes the openPMD file (e.g. the unit system adopted internally
+by a physics simulation code), in order to avoid a reduction in performance
+associated with unit conversion during the write process.
+
+However, for each dataset and attribute, it is **very strongly** recommended to
+provide the conversion factor from the chosen unit system to the International
+System of Units (SI). (See the description of the attributes `timeUnitSI`, `gridUnitSI` and `unitSI`, in the present document.) This allows
```
ax3l: line length is broken :)

STANDARD.md
```diff
@@ -584,19 +590,37 @@ patch order:
...
 
-While this standard does not impose any unit system on the data that is stored
-itself, it still requires a common interface to convert one unit system to
-another.
-
-To allow scaling data without reformatting it during the write process we
-provide a unit conversation factor, often called `unitSI` in the document,
-to transform it to a corresponding quantity in the International System of
-Units (SI).
+All datasets and attributes can be written in arbitrary units, in the openPMD
```
ax3l: superfluous `,`?

ax3l: Also I would avoid "the openPMD file" in new text. openPMD is not bound to files. Just write `in openPMD` or skip it at all - openPMD is what these docs are about :)

STANDARD.md
```diff
@@ -584,19 +590,37 @@ patch order:
...
+file. Typically, this will be the unit system adopted internally by the
+software that writes the openPMD file (e.g. the unit system adopted internally
```
ax3l: software that writes ~~the~~ openPMD ~~file~~

STANDARD.md
```diff
@@ -584,19 +590,37 @@ patch order:
...
+by a physics simulation code), in order to avoid a reduction in performance
+associated with unit conversion during the write process.
+
+However, for each dataset and attribute, it is **very strongly** recommended to
+provide the conversion factor from the chosen unit system to the International
+System of Units (SI). (See the description of the attributes `timeUnitSI`, `gridUnitSI` and `unitSI`, in the present document.) This allows
+the reading softwares to convert the datasets to units that any user can
+understand, without requiring them to know the chosen units system.
```
ax3l: unit~~s~~ system?

STANDARD.md
```diff
@@ -584,19 +590,37 @@ patch order:
...
+
+(Note that, for human readable output, it is possible to add the actual string
```
ax3l: I would not add this note here since it's confusing with out concept of automatically deriving actual units. We might rather add a reference here to `unitDimension` and add advice to implementers and users therein.

ax3l (commented):
Looks good, just a few inline comments :)

STANDARD.md
RemiLehe: Ok, I just removed this note altogether.

RemiLehe:
@ax3l Thanks for the review ; I just updated the PR.

ax3l (approved):

PR #192, opened 2018-04-03
Remove proton/neutron Number

### LINKED ISSUE #191: ED-PIC: Remove proton/neutron Number?
With the new `speciesType` attribute in an extension #180, we can remove (and auto-update) the optional `protonNumber` and `neutronNumber` attribute in the `ED-PIC` extension.

RemiLehe:
This removes the proton/neutron number from the `ED-PIC` extension, since it is now possible to specify it with `SpeciesType`.

*Implements issue:* Close #191

## Affected Components

- `EXT_ED-PIC`

## Logic Changes

Expressing ions in `ED-PIC` can now be done by using the `SpeciesType` extension.

## Writer Changes

Data writers that were writing the proton/neutron number should now use the `SpeciesType` extension.

This might require a lookup of an *element name* in the periodic table for the naming. Previously, one could just define two numbers without knowing how an ion/isotope is abbreviated in the periodic table.

## Reader Changes

I am not aware of any reader that currently checks this attribute, but in theory data readers should now look at the `SpeciesType` information.

- [x] `openPMD-validator`: remove check in `ED-PIC` https://github.com/openPMD/openPMD-validator/pull/39

## Data Converter

Data converters should switch to using the `SpeciesType`.

- [ ] https://github.com/openPMD/openPMD-updater

### DIFF
--- a/EXT_ED-PIC.md
+++ b/EXT_ED-PIC.md
@@ -226,6 +226,10 @@ Particle Records (Macroparticles)
 The following additional attributes are defined in this extension.
 The individual requirement is given in `scope`.
 
+Note that the attributes below to not include information about the species
+type (e.g. electrons, Helium 4, etc.). Please use the [SpeciesType
+extension](EXT_SpeciesType.md) in order to include this type of information.
+
   - `particleShape`
     - type: *(floatX)*
     - scope: *required*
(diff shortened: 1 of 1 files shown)


ax3l:
> I am not aware of any reader that currently checks this attribute

We do in manual post-processing from time to time.

> Does the openPMD validator check the proton number?

Yes, it does check its type. We need to remove these two lines: https://github.com/openPMD/openPMD-validator/pull/39

> Data converters should switch to using the SpeciesType.

I would like to hold this PR until we have a [data converter](https://github.com/openPMD/openPMD-updater) PR for it. This way of working we will only merge changes to the standard that we can also convert to :)

(The data conversion will need to implement a periodic table lookup, that's fun! :-p )

EXT_ED-PIC.md
```diff
@@ -226,6 +226,10 @@ Particle Records (Macroparticles)
 The following additional attributes are defined in this extension.
 The individual requirement is given in `scope`.
 
+Note that the attributes below to not include information about the species
```
ax3l, line 229: to -> do
```suggestion
Note that the attributes below do not include information about the species
```

ax3l (commented):
Just a minor typo.

Issue #187, opened 2018-02-26
Put in **Definitions** section in the Standard


DavidSagan:
I propose to put in a **Definitions** section in the standard to make it easier for people to look up what things mean. 

Some quick notes:

1) If a **Definitions** section is put in there would have to be some cleanup in other parts of the Standard.
2) Do we want "dataset" instead of "data set" since what is being referred to here is not a generic set of data but a particular construct?
3) The `Naming conventions` for records should be merged into the **Definitions** section.

Here is a quick draft to open the discussion:

- **Group**: A **group** is a container structure containing a set of zero or more **attributes**, a set of zero or more **groups** (which can be called **sub-groups**), and a set of zero or more **datasets**. Note: In HDF5, these are also called **groups**.

- **Dataset**: A **dataset** is a structure that contains a set of zero or more **attributes** and a
data aray (which may be multidimensional).  Note: In HDF5, these are also called **datasets**.

- **Record**: A **record** is a **group** (without any sub-groups) or a **dataset** that contains data on a physical quantity like particle charge or electric field. There are three types of **records**:
    - **scalar records** hold scalar quantity
    values (like particle charge). If all the particles have the same charge, the value of the charge
    is stored as an attribute of the **record** and there is no associated data array. That is, the
    **record** is a **group**. If the particles have differing charges, the values are stored in an array
    of the **scalar record**. In this case the **scalar record** is a **dataset**.
    - **vector records** hold a set of **datasets**. In this case the **record** is a **group**.
        - Example: A **record** named **E** for holding electric field values may have three datasets holding the components of the field named **E/x**, **E/y**, and **E/z**.
    - **tensor record** hold tensor quantities.

  Records with only a scalar component are stored in a data set with the same name as the record. Vector and tensor records shall be represented component-wise as a *collection of individual scalar data sets* using a common sub-group that is equal to the record name.

  We refer to the scalar record itself and the vector sub-group as `record`, to the data sets in the vector sub-group as `components`. For scalar records, the `record` is the `component` (and vice versa).

- **Attribute**: An **attribute** is a variable associated with a **group** along with a value. Example: **snapshotPath** is a string variable associated with the root **/** group.

- **Particle Root Group**: A **Particle Root Group** is a group for specifying a set of particles. There can be multiple **particle root groups** in a data file. **Particle root groups** are subgroups of the `particlesPath` group.

[...]

Issue #129, opened 2016-05-19
Simply `axisLabels` (without dataOrder)


ax3l:
Related to #125: the naming of the indexes to axes of matrices, aka `axisLabels`, should simply be ordered from fastest to slowest varying index when accessing its memory contiguously.
### Example Matrix for `axisLabels = "x;y;z"`

that order means, that the _flattened_ matrix has `x` as its _fastest varying index_ and `y` as the _slower varying index_ and `z` as the _slowest varying index_:

**Fortran** (_first_ index varies fastest in contiguous memory; ["column-major"](http://docs.scipy.org/doc/numpy-1.10.0/glossary.html#term-column-major))

``` Fortran90
real(kind=8), dimension(size_x, size_y, size_z) :: A

! efficient write access
do z = 1, size_z
   do y = 1, size_y
      do x = 1, size_x
         a(x,y,z) = 1.0
      end do
   end do
end do
```

**C/C++** (_last_ index varies fastest in contiguous memory; ["row-major"](http://docs.scipy.org/doc/numpy-1.10.0/glossary.html#term-row-major))

``` C++
double A[size_z][size_y][size_x];

// efficient write access
for( size_t z = 0; z < size_z; ++z )
    for( size_t y = 0; y < size_y; ++y )
        for( size_t x = 0; x < size_x; ++x )
            A[z][y][x] = 1.0;
```

**Python numpy.ndarray** (_last_ index varies fastest in contiguous memory; ["row-major"](http://docs.scipy.org/doc/numpy-1.10.0/glossary.html#term-row-major))

``` python
# Python arrays are C-like arrays as long as `order` is not changed
# See:
#   http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.ndarray.html#numpy.ndarray
#   http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.ndarray.flatten.html#numpy.ndarray.flatten
#   http://docs.scipy.org/doc/numpy-1.10.0/glossary.html#term-row-major
A = np.ones(shape=(size_z,size_y,size_x), dtype=np.float64)

# one should not loop in such a way in python but if one would,
# here is the efficient index access order accessing all elements
# contiguously:
for z in np.arange(size_z, dtype=np.uint):
    for y in np.arange(size_y, dtype=np.uint):
        for x in np.arange(size_x, dtype=np.uint):
            A[z,y,x] = 1.0
```

[...]

Issue #125, opened 2016-04-26
`dataOrder` is superfluous


ax3l:
By defining things like `axisLabel` are ordered by fastest to slowest varying index one can probably omit `dataOrder`.

needs a bit more evaluation but should be true.

DavidSagan:
Yes I think `dataOrder` should be removed or at least made optional.

anokfireball:
After developing two readers/writers that deal with openPMD data and having read the standard numerous times, ```dataOrder``` and ```axisLabels``` are two attributes I **always** have to look up.

I presume ```dataOrder``` was introduced to handle file formats that do not explicitly store which dimensions are fastest. Considering the major formats:
- HDF5 stores row-major only (their Fortran wrapper transposes accordingly)
- ADIOS stores the native order, offers [functionality](https://github.com/ornladios/ADIOS/blob/master/src/core/common_read.c#L614) to determine the order, does not transpose out of the box
- NetCDF stores row-major only (their Fortran wrapper transposes accordingly)
- SDF stores row-major only ([xyz](https://en.wikipedia.org/wiki/Chemical_table_file))
- VTK stores row-major only ([tensors](https://www.vtk.org/wp-content/uploads/2015/04/file-formats.pdf))

... makes the order implicit in most formats and detectable in the remaining one.

I propose to make ```dataOrder``` optional in 2.0. ```axisLabels``` should reflect the natural order, depending on the order in the file. If a file format supports both orders, but does not allow to query the order, ```dataOrder``` should be supplied.

Also, the description of ```axisLabels``` in the standard should explicitly state whether the order is from slowest to fastest or from fastest to slowest.

ax3l:
Excellent, I will put you both on the review of that change then.

I would actually remove `dataOrder` completely and describe attributes such as `axisLabel` more precisely. Making it optional will not lift the confusion.

ax3l:
Resolved with #194 

See also: https://github.com/openPMD/openPMD-standard/issues/235#issuecomment-2130075297

PR #193, opened 2018-04-23
GridSpacing: Per-Component UnitSI & UnitDimension

### LINKED ISSUE #122: GridSpacing: Per-Component UnitSI & UnitDimension
The `gridSpacing` should get a `unitSI` and a `unitDimension` for each of its components.

This allows to describe grid such as `x, ux` or `x,y,t` (although the "time" definition per iteration would then be "for the first row").

RemiLehe:
*Implements issue:* Close #122 

## Description

This allows meshes to have axes with different dimensions (e.g. for phase space data where one axis is a position, and the second axis is a momentum.)

The attributes `gridUnitDimension` and `gridUnitSI` are now `N` times longer, where `N` is the dimension of the mesh.
In particular `gridUnitDimension` is a 1D array of length `7*N`. This was preferred over a 2D array of dimension `(N,7)` since it seems that `h5py` does not support multidimensional arrays as attributes. 

## Affected Components

- `base`: All mesh data

## Writer Changes

Data writers should now write the longer array for `gridUnitSI` and add `gridUnitDimension`.

**Note:**  We could make this backward compatible by saying that, if the writer wrote the previous short-array format, then this applies to each axis... @ax3l Is this worth it or too confusing. -> too confusing, we have the updater for that :)

- `openPMD-api` (upcoming): https://github.com/openPMD/openPMD-api/...

## Reader Changes

Data readers should now extract the dimension of each axis.

- `openPMD-validator`: https://github.com/openPMD/openPMD-validator/pull/40

## Data Converter

For existing attributes should be duplicated `N` times, in order to convert to the new format.

- [x] https://github.com/openPMD/openPMD-updater/pull/7

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -95,7 +95,7 @@ Each file's *root* group (path `/`) must at least contain the attributes:
       present standard (e.g. fields on an unstructured mesh),
       this can be done be storing this data within a path that *is not*
       of the form given by `basePath` (e.g. `/extra_data`). In this
-      way, the openPMD parsing tools will not parse this additional data.
+      way, the openPMD data readers will not parse this additional data.
 
 The following attribute is *optional* in each each file's *root* group
 (path `/`) and indicates if a file also follows an openPMD extension
@@ -425,7 +425,7 @@ meshes):
 
   - `gridUnitSI`
     - type: 1-dimensional array containing N *(float64 / REAL8)*
-            elements, where N is the number of dimensions in the simulation
+            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from fastest to slowest varying index of the described mesh.
     - description: unit-conversion factor to multiply each value in
                    `gridSpacing` and `gridGlobalOffset`, in order to convert
                    from simulation units to SI units
@@ -433,7 +433,7 @@ meshes):
 
   - `gridUnitDimension`
     - type: array of 7 N *(float64 / REAL8)*
-            elements, where N is the number of dimensions in the simulation
+            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from fastest to slowest varying index of the described mesh.
     - description: powers of the 7 base measures characterizing the
             grid axes's dimensions (length L, mass M, time T,
             electric current I, thermodynamic temperature theta,
@@ -441,12 +441,13 @@ meshes):
     - note: this is similar to `unitDimension`, but applies to each axis
         The 7 numbers characterizing the dimension of an axis are stored
         contiguously in this 1D array.
-    - examples:
+    - examples (with `L`, `M` and `T` as defined in `unitDimension`)
         - For a 2D spatial grid (`L=1`), store array
         `(1., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0.)`
-        - For a 2D phase space (Unit "m" along the first axis
-        and "kg.m.s-1" along the second axis), store array
+        - For a 2D phase space (Dimension "L" along the first axis
+        and "L*M/T" along the second axis), store array
         `(1., 0., 0., 0., 0., 0., 0., 1., 1., -1., 0., 0., 0., 0. )`
+
 
 The following attributes must be stored with each `scalar record` and each
 *component* of a `vector record`:
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -431,7 +431,7 @@ on Unit Systems and Dimensionality, further below):
 
   - `gridUnitSI`
     - type: 1-dimensional array containing N *(float64 / REAL8)*
-            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from fastest to slowest varying index of the described mesh.
+            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from slowest to fastest varying index of the described mesh.
     - description: unit-conversion factor to multiply each value in
                    `gridSpacing` and `gridGlobalOffset`, in order to convert
                    from simulation units to SI units
@@ -439,7 +439,7 @@ on Unit Systems and Dimensionality, further below):
 
   - `gridUnitDimension`
     - type: array of 7 N *(float64 / REAL8)*
-            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from fastest to slowest varying index of the described mesh.
+            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from slowest to fastest varying index of the described mesh.
     - description: powers of the 7 base measures characterizing the
             grid axes's dimensions (length L, mass M, time T,
             electric current I, thermodynamic temperature theta,
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -431,15 +431,17 @@ on Unit Systems and Dimensionality, further below):
 
   - `gridUnitSI`
     - type: 1-dimensional array containing N *(float64 / REAL8)*
-            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from slowest to fastest varying index of the described mesh.
+            elements, where N is the number of dimensions in the simulation.
+            The order of the N values must be identical to the axes in `axisLabels`.
     - description: unit-conversion factor to multiply each value in
                    `gridSpacing` and `gridGlobalOffset`, in order to convert
                    from simulation units to SI units
     - example: `(1.0e-9, 1.0e-9, 1.0e-6)`
 
   - `gridUnitDimension`
-    - type: array of 7 N *(float64 / REAL8)*
-            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from slowest to fastest varying index of the described mesh.
+    - type: 1-dimensional array of 7 N *(float64 / REAL8)*
+            elements, where N is the number of dimensions in the simulation.
+            The order of the N 7-value arrays must be identical to the axes in `axisLabels`.
     - description: powers of the 7 base measures characterizing the
             grid axes's dimensions (length L, mass M, time T,
             electric current I, thermodynamic temperature theta,
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -95,7 +95,7 @@ Each file's *root* group (path `/`) must at least contain the attributes:
       present standard (e.g. fields on an unstructured mesh),
       this can be done be storing this data within a path that *is not*
       of the form given by `basePath` (e.g. `/extra_data`). In this
-      way, the openPMD parsing tools will not parse this additional data. 
+      way, the openPMD parsing tools will not parse this additional data.
 
 The following attribute is *optional* in each each file's *root* group
 (path `/`) and indicates if a file also follows an openPMD extension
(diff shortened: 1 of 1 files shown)


ax3l:
> Note: We could make this backward compatible by saying that, if the writer wrote the previous short-array format, then this applies to each axis...

Very good point. I think it's too special to allow this short-cut which just adds extra-checks in readers and writers. So let's not allow to just write it once as it was allowed before.

STANDARD.md
```diff
@@ -95,7 +95,7 @@ Each file's *root* group (path `/`) must at least contain the attributes:
       present standard (e.g. fields on an unstructured mesh),
       this can be done be storing this data within a path that *is not*
       of the form given by `basePath` (e.g. `/extra_data`). In this
-      way, the openPMD parsing tools will not parse this additional data. 
+      way, the openPMD parsing tools will not parse this additional data.
```
ax3l: while we are on it: it's not only the *openPMD parsing tools* but general *openPMD data readers*

STANDARD.md
```diff
@@ -424,11 +424,29 @@ meshes):
...
+            grid axes's dimensions (length L, mass M, time T,
+            electric current I, thermodynamic temperature theta,
+            amount of substance N, luminous intensity J)
+    - note: this is similar to `unitDimension`, but applies to each axis
+        The 7 numbers characterizing the dimension of an axis are stored
+        contiguously in this 1D array.
+    - examples:
+        - For a 2D spatial grid (`L=1`), store array
+        `(1., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0.)`
+        - For a 2D phase space (Unit "m" along the first axis
```
ax3l: Maybe we better stay with `L` and `M * L / T` as defined in `unitDimension`?

STANDARD.md
```diff
@@ -424,11 +424,29 @@ meshes):
     - example: `(0.0, 100.0, 0.0)` or `(0.5, 0.5, 0.5)`
 
   - `gridUnitSI`
-    - type: *(float64 / REAL8)*
+    - type: 1-dimensional array containing N *(float64 / REAL8)*
+            elements, where N is the number of dimensions in the simulation
```
ax3l: This is related to #125 & #129 (PR in #194): if we flatten the array we have to specify *how*.
I would propose to use the wording from #129 as such:
`, dimensions shall be ordered from fastest to slowest varying index [of the described mesh].`

STANDARD.md
```diff
@@ -424,11 +424,29 @@ meshes):
...
     - description: unit-conversion factor to multiply each value in
                    `gridSpacing` and `gridGlobalOffset`, in order to convert
                    from simulation units to SI units
-    - example: `1.0e-9`
+    - example: `(1.0e-9, 1.0e-9, 1.0e-6)`
+
+  - `gridUnitDimension`
+    - type: array of 7 N *(float64 / REAL8)*
+            elements, where N is the number of dimensions in the simulation
```
ax3l: we need to add the same comment with order as above for flattening

STANDARD.md
```diff
@@ -424,11 +424,29 @@ meshes):
...
+    - type: 1-dimensional array containing N *(float64 / REAL8)*
+            elements, where N is the number of dimensions in the simulation
     - description: unit-conversion factor to multiply each value in
                    `gridSpacing` and `gridGlobalOffset`, in order to convert
                    from simulation units to SI units
-    - example: `1.0e-9`
+    - example: `(1.0e-9, 1.0e-9, 1.0e-6)`
+
+  - `gridUnitDimension`
+    - type: array of 7 N *(float64 / REAL8)*
```
ax3l: I would write `N arrays of 7 *(float64 / REAL8)*`

STANDARD.md
ax3l: update: we probably will go from `slowest to fastest` in #194

STANDARD.md
RemiLehe: Ok, that's one possibility.
However, the thing is that it is stored as a single 1D array.

STANDARD.md
```diff
@@ -425,29 +425,30 @@ meshes):
 
   - `gridUnitSI`
     - type: 1-dimensional array containing N *(float64 / REAL8)*
-            elements, where N is the number of dimensions in the simulation
+            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from fastest to slowest varying index of the described mesh.
```
ax3l: will be inverted

STANDARD.md
ChristopherMayes: BTW, I see a few REAL8 mentioned, but this word isn't actually standard (Neither is real(8)) in Fortran. Fortran 2008 provides standard kinds: REAL32, REAL64, REAL128
http://fortranwiki.org/fortran/show/Real+precision

STANDARD.md
```diff
@@ -424,11 +424,30 @@ meshes):
...
     - description: unit-conversion factor to multiply each value in
                    `gridSpacing` and `gridGlobalOffset`, in order to convert
                    from simulation units to SI units
-    - example: `1.0e-9`
+    - example: `(1.0e-9, 1.0e-9, 1.0e-6)`
+
+  - `gridUnitDimension`
+    - type: array of 7 N *(float64 / REAL8)*
+            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from fastest to slowest varying index of the described mesh.
```
ax3l: slowest to fastest

STANDARD.md
```diff
@@ -424,11 +424,30 @@ meshes):
     - example: `(0.0, 100.0, 0.0)` or `(0.5, 0.5, 0.5)`
 
   - `gridUnitSI`
-    - type: *(float64 / REAL8)*
+    - type: 1-dimensional array containing N *(float64 / REAL8)*
+            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from fastest to slowest varying index of the described mesh.
```
ax3l: slowest to fastest

ax3l (changes requested):
Annotated the changed index order which is introduced in  #194

STANDARD.md
ax3l: Huh, very good point. This is also something for #187 (please cross-post therein again)

ax3l (approved):
Changes look great!

Let us just add the PR to the converter before we merge.

PrometheusPi:
@ax3l Does this require changing how we write hdf5 data in the phase space and radiation plugin in [PIConGPU](https://github.com/ComputationalRadiationPhysics/picongpu)?

ax3l:
@PrometheusPi there is a whole bunch of changes that are [changing openPMD](https://github.com/openPMD/openPMD-standard/issues?q=is%3Aopen+label%3A%22major+change%22+project%3AopenPMD%2FopenPMD-standard%2F3) in a incompatible way in 2.0.0. This is one of them and will enable that we can describe phase space and radiation plugin in valid openPMD.

When we have all changes incorporated, [a changelog](https://github.com/openPMD/openPMD-standard/blob/1.1.0/CHANGELOG.md#110) will summarize all of them again. Also, we can then use [openPMD-api](https://github.com/openPMD/openPMD-api) instead of libSplash to write from all PIConGPU plugins.

We will then update all places of PIConGPU and then add openPMD to phase space and radiation plugin.

STANDARD.md
ax3l: >  However, the thing is that it is stored as a single 1D array.

Ah I see, ok with me.

STANDARD.md
```diff
@@ -430,11 +430,30 @@ In addition, the following attribute is *recommended* (see the section
...
+            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from slowest to fastest varying index of the described mesh.
     - description: unit-conversion factor to multiply each value in
                    `gridSpacing` and `gridGlobalOffset`, in order to convert
                    from simulation units to SI units
-    - example: `1.0e-9`
+    - example: `(1.0e-9, 1.0e-9, 1.0e-6)`
+
+  - `gridUnitDimension`
+    - type: array of 7 N *(float64 / REAL8)*
+            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from slowest to fastest varying index of the described mesh.
```
ax3l: Overlong line. Maybe put "Dimensions shall be..." on the next line (indented to same block).

STANDARD.md
DavidSagan: My thought would be to not even mandate any precision (ie just say "real"). For example, why should the standard disallow the use of real4?

STANDARD.md
ax3l: I would try to make this more clear how the arrays are stored by writing:

"The order of the N 7-value arrays must be identical to the axes in `axisLabels`."

This makes it consistent with the other attributes above that have the same order and uses the same wording, also consistently with #194.
```suggestion
            elements, where N is the number of dimensions in the simulation.
            The order of the N 7-value arrays must be identical to the axes in `axisLabels`.
```

STANDARD.md
ax3l: The unit dimension attribute is not a record but just describing the dimensionality ("unit"). It introduces no problem keeping it fixed and makes parsing very simple.

ax3l: @RemiLehe pls replace `array` with `1-dimensional array` for consistency.
```suggestion
    - type: 1-dimensional array of 7 N *(float64 / REAL8)*
```

STANDARD.md
```diff
@@ -430,11 +430,30 @@ In addition, the following attribute is *recommended* (see the section
 on Unit Systems and Dimensionality, further below):
 
   - `gridUnitSI`
-    - type: *(float64 / REAL8)*
+    - type: 1-dimensional array containing N *(float64 / REAL8)*
+            elements, where N is the number of dimensions in the simulation. Dimensions shall be ordered from slowest to fastest varying index of the described mesh.
```
ax3l:
```suggestion
            elements, where N is the number of dimensions in the simulation.
            The order of the N values must be identical to the axes in `axisLabels`.
```

ax3l (approved):

STANDARD.md
```diff
@@ -430,11 +430,32 @@ In addition, the following attribute is *recommended* (see the section
 on Unit Systems and Dimensionality, further below):
 
   - `gridUnitSI`
-    - type: *(float64 / REAL8)*
+    - type: 1-dimensional array containing N *(float64 / REAL8)*
```
franzpoeschel, line 433: Note: This does not extend, but instead changes the definition of `gridUnitSI`. According to this wording, writing a single scalar to be used for all axes will no longer be supported for 2.0, meaning that all writers need to change this.
Do we additionally want to allow scalars here in 2.0, or should writers migrate?

Currently implementing this by printing a warning for legacy-defined scalar values for `gridUnitSI` in https://github.com/openPMD/openPMD-api/pull/1534/:
```cpp
Mesh &Mesh::setGridUnitSI(double gusi)
{
    setAttribute("gridUnitSI", gusi);
    if (auto standard = IOHandler()->m_standard;
        standard >= OpenpmdStandard::v_2_0_0)
    {
        std::cerr << "[Mesh::setGridUnitSI] Warning: Setting a scalar "
                     "`gridUnitSI` in a file with openPMD version '" +
                std::string(auxiliary::formatStandard(standard)) +
                "'. Consider specifying a vector instead in order to "
                "specify the gridUnitSI per axis (ref.: "
                "https://github.com/openPMD/openPMD-standard/pull/193).\n";
    }
    return *this;
}
```

franzpoeschel:
Implemented in the API with https://github.com/openPMD/openPMD-api/pull/1534

PR #194, opened 2018-04-24
Mesh: Remove Data Order

ax3l:
The data order for indices arises naturally from its flattened memory layout.
This also holds true for logical access to the data.

This affects the *base standard* attributes:
- `axisLabels`
- `gridSpacing`
- `gridGlobalOffset`
- `shape` of constant record components

and the *ED-PIC* extension attributes:
- `fieldBoundary`
- `particleBoundary`

*Implements issues:* #125 and #129

## Description

`dataOrder` is superfluous.

By defining 1D-flattened dimensions like `axisLabel` are ordered by slowest to fastest varying index one can omit `dataOrder`. Slowest to fastest varying index here only means the logical flattened memory layout that one assumes for this data format.

This is possible since one does not care *how* data was written but just about *how to label/name indices* when reading data. E.g., one would increase from fastest to slowest dimension for all flattened descriptions over dimensions to read data efficiently without seeking.

*Let's have a C-like to Fortran-like example*

A C-matrix `A[i,j,k]` is written with `i` index standing for `"x"`, `j` for `"y"` and `k` for `"z"`. The associated `axisLabels` is in the index order k-j-i (slowest is first) which is `("x", "y", "z")`.

A Fortran-reader reads this matrix as `M(a,b,c)` with the index data order a-b-c (`c` is varies slowest). The `axisLabels` attribute reads `("x", "y", "z")`. So one will assign the fastest varying index `c` the label `"z"` , `b` the label `"y"` and the fastest varying index `a` an `"x"`. Index order and index label/name conserved - hurray! :sparkles: 

## Affected Components

- `base`
- `EXT: ED-PIC`

## Logic Changes

Before one had to to write the access-index order of the writing code. Now one writes the labels in the order "slowest-to-fastest varying index".

This does not change C-style, but Fortran-style access.

## Writer Changes

Yes, Fortran-style writers need to invert their order now since the fastest varying index is the last.
C-style writers (C/C++/Python/Java/...) can just omit the `dataOrder` attribute now.

*Unaffected besides the now removed attribute:*

- `PIConGPU` (C++): https://github.com/ComputationalRadiationPhysics/picongpu/...
- `Warp` (Python-Writer): ...
- `FBPIC` (Python): ...
- `Smilei` (C++): https://github.com/SmileiPIC/Smilei
- `SIMEX Platform` (Python): https://github.com/eucall-software/simex_platform
- `openPMD-api` (C++): https://github.com/openPMD/openPMD-api/...

Fortran/Matlab/R/Julia codes: ...

## Reader Changes

Yes, *all readers* can now just label/name flattened dimension accesses to e.g. labels of indices in the order "slowest-to-fastest index".

- [x] `openPMD-validator`: https://github.com/openPMD/openPMD-validator/pull/42
- `postpic`: https://github.com/skuschel/postpic
- `openPMD-api`: https://github.com/openPMD/openPMD-api/...

## Data Converter

Mesh records that declare `dataOrder='F'` need to invert the attributes mentioned above.

The unused `dataOrder` attribute should be removed as well.

- [x] https://github.com/openPMD/openPMD-updater/pull/4

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -366,41 +366,29 @@ meshes):
                         the *imaginary* part)
                        ![definition of imaginary part](img/cylindrical.png)
 
-  - `dataOrder`
-    - type: *(string)*
-    - description: used for the reading of 1-dimensional arrays of N elements,
-                   where N is the number of dimensions in the simulation;
-                   these should be in the ordering of variation for the indexes
-                   for matrices as defined by the index-operator (`[...][...]`)
-                   of the writing code; can be omitted for 1D records
-    - rationale: in Fortran, ordering of matrices is linearized in memory in
-                 column-major order whereas in C it is row-major; due to that
-                 the index-operators in Fortran and C operate in exactly
-                 opposite order;
-                 supported file-formats keep the order of the writing code's
-                 record components; we still need to store additional
-                 information such as `axisLabels` in a defined order
-    - allowed values:
-      - `C` if data is written by C or a C-like language such as C++, Python,
-        Java
-      - `F` if data is written by a Fortran code
-
   - `axisLabels`
     - type: 1-dimensional array containing N *(string)*
             elements, where N is the number of dimensions in the simulation
-    - description: ordering of the labels for the `geometry` of the mesh
-    - advice to implementors: in the ordering of variation for the indexes for
-                              matrices as defined by the index-operator
-                              (`[...][...]`) of the writing code
-    - advice to implementors: on read, query the record's `dataOrder` to get the
-                              information if you need to invert the access to
-                              `axisLabels` (and other attributes that use the
-                              same definition)
+    - description: this attribute assigns human-readible labels for the
+                   indices `i`, `j`, `k`, etc. denoting the axes of a mesh
+                   `A_{i,j,k}`
+    - advice to implementors: dimensions shall be ordered from slowest to
+                              fastest varying index when accessing the mesh
+                              contiguously (as 1D flattened logical memory)
+    - advice to implementors: if you access a ND array in C-like languages,
+                              a matrix `A[i,j,k]` will have its first index
+                              as the slowest varying index (e.g. `i`);
+    -                         if you access a ND array Fortran-like,
+                              a matrix `A(i,j,k)` will have its last index
+                              as the slowest varying index (e.g. `k`);
     - examples:
-      - 3D `cartesian` C-style `A[z,y,x]` write: `("z", "y", "x")` and `dataOrder='C'`
-      - 2D `cartesian` C-style `A[y,x]` write: `("y", "x")` and `dataOrder='C'`
-      - 2D `cartesian` Fortran-style `A[x,y]` write: `("x", "y")` and `dataOrder='F'`
-      - `thetaMode` Fortran-style `A[r,z]` write: `("r", "z")` and `dataOrder='F'`
+      - 3D `cartesian` mesh accessed in C-like as `A[z,y,x]` will have `z` as
+        its slowest varying index name and `axisLabels`: `("z", "y", "x")`
+      - 3D `cartesian` mesh accessed in C-like as `A[x,y,z]` will have `x` as
+        its slowest varying index name and `axisLabels`: `("x", "y", "z")`
+      - 2D `cartesian` mesh accessed in Fortran-like as `A(x,y)` will have `y` as
+        its slowest varying index name and `axisLabels`: `("y", "x")`
+      - `thetaMode` accessed Fortran-like `A(r,z)`, `axisLabels`: `("z", "r")`
 
   - `gridSpacing`
     - type: 1-dimensional array containing N *(floatX)*


STANDARD.md
```diff
@@ -366,41 +366,28 @@ meshes):
...
-    - advice to implementors: in the ordering of variation for the indexes for
-                              matrices as defined by the index-operator
-                              (`[...][...]`) of the writing code
-    - advice to implementors: on read, query the record's `dataOrder` to get the
-                              information if you need to invert the access to
-                              `axisLabels` (and other attributes that use the
-                              same definition)
+    - description: this attribute assigns for a matrix `A_{i,j,k}`
+                   human-readible namings for the indices `i`, `j` and `k`
+    - advice to implementors: dimensions shall be ordered from fastest to
```
anokfireball: This clarification really makes the PR valuable.

anokfireball (approved):
Simplifying the standard by integrating ```dataOrder``` into ```axisLabels``` improves understandability a lot. The user will have to be aware of this order, but now ambiguities that reference ```axisLabels``` (but not ```dataOrder```) are resolved.

mccoys:
If I understand correctly, ~~C-order~~ *F-order* data will not be actually modified, but only the following attributes must have the order of their elements reversed *when writing from Fortran*?

*    axisLabels
*   gridSpacing
*    gridGlobalOffset
*    shape of constant record components
*    fieldBoundary
*    particleBoundary

ax3l:
@mccoys ~~from a C/C++/Python *writer* perspective (Smilei), yes just invert the order of those attributes when writing openPMD 2.0~~

update: we changed the order from originally proposed *fastest-to-slowest* to *slowest-to-fastest*.

From a Fortran *writer* perspective, yes just invert the order of those attributes when writing openPMD 2.0
From a C/C++/Python *writer* perspective everything stays the same.

ax3l:
There is also an alternative solution to this change, *ordering from slowest-to-fastest* varying index.

With that the existing Fortran-written files would need to change their index and we would generally have everything in C-like order. Due to the dominance of Py/C++ in our code bases, one could decide for that solution and have more consistency with e.g. numpy attributes >:-)

STANDARD.md
```diff
@@ -366,41 +366,28 @@ meshes):
...
             elements, where N is the number of dimensions in the simulation
-    - description: ordering of the labels for the `geometry` of the mesh
-    - advice to implementors: in the ordering of variation for the indexes for
-                              matrices as defined by the index-operator
-                              (`[...][...]`) of the writing code
-    - advice to implementors: on read, query the record's `dataOrder` to get the
-                              information if you need to invert the access to
-                              `axisLabels` (and other attributes that use the
-                              same definition)
+    - description: this attribute assigns for a matrix `A_{i,j,k}`
```
RemiLehe: Could you clarify the meaning of `assigns for`, or maybe use some other phrasing?

RemiLehe: One possibility is 
```
- description: human-reading labels for the `N` axes of the mesh,
ordered *from fastest to slowest varying index when accessing the
mesh contiguously (as 1D flattened logical memory)`

**Review by RemiLehe — CHANGES_REQUESTED** [review]

File: STANDARD.md (hunk 2 above)
Comment (ax3l) [reply STANDARD.md]: I think I wanted to write `assigns human-readible namings for the indices ... to a matrix ...`

**Comment by ax3l** [comment]
@RemiLehe is it ok for you as well if I [change the order](https://github.com/openPMD/openPMD-standard/pull/194#issuecomment-389117002) to *slowest to fastest* varying index? :)

VC: ok with all.

**Comment by ax3l** [comment]
all right, I updated the PR and PR description! :)

File: STANDARD.md [hunk 3]
```diff
@@ -366,41 +366,29 @@ meshes):
...
+                   indices `i`, `j`, `k`, etc. denoting the axes of a mesh
+                   `A_{i,j,k}`
+    - advice to implementors: dimensions shall be ordered from slowest to
+                              fastest varying index when accessing the mesh
+                              contiguously (as 1D flattened logical memory)
+    - advice to implementors: if you access a ND array in C-like languages,
+                              a matrix `A[i,j,k]` will have its first index
+                              as the slowest varying index (e.g. `i`);
+    -                         if you access a ND array Fortran-like,
+                              a matrix `A[i,j,k]` will have its last index
```
ax3l: Fortran spells arrays `A(i,j,k)`

STANDARD.md
```diff
@@ -366,41 +366,29 @@ meshes):
...
     - examples:
-      - 3D `cartesian` C-style `A[z,y,x]` write: `("z", "y", "x")` and `dataOrder='C'`
-      - 2D `cartesian` C-style `A[y,x]` write: `("y", "x")` and `dataOrder='C'`
-      - 2D `cartesian` Fortran-style `A[x,y]` write: `("x", "y")` and `dataOrder='F'`
-      - `thetaMode` Fortran-style `A[r,z]` write: `("r", "z")` and `dataOrder='F'`
+      - 3D `cartesian` mesh accessed in C-like as `A[z,y,x]` will have `z` as
+        its slowest varying index name and `axisLabels`: `("z", "y", "x")`
+      - 3D `cartesian` mesh accessed in C-like as `A[x,y,z]` will have `x` as
+        its slowest varying index name and `axisLabels`: `("x", "y", "z")`
+      - 2D `cartesian` mesh accessed in Fortran-like as `A[x,y]` will have `y` as
```
ax3l: `A(x,y)`

STANDARD.md
```diff
@@ -366,41 +366,29 @@ meshes):
...
+        its slowest varying index name and `axisLabels`: `("y", "x")`
+      - `thetaMode` accessed Fortran-like `A[r,z]`, `axisLabels`: `("z", "r")`
```
ax3l: `A(r,z)`

RemiLehe (approved):

ax3l:
@RemiLehe this PR should be ready to merge. Validator and Updater are up-to-date now :)

Issue #161, opened 2017-12-10
Make "time", "dt" and "timeUnitSI" optional in basePath


DavidSagan:
Since "time", "dt" and "timeUnitSI" do not make sense, say, when particle positions are recorded at constant longitudinal position. These quantities should be made optional.

@ax3l: Please put this in the 2.0 project.

ax3l:
Thank you for the issue!

Can you elaborate on "particle positions are recorded at constant longitudinal position". Wouldn't these positions be stored over a iteration (snapshot) that can be referred to a time again? You mean e.g. a track of a particle within a single iteration (snapshot) group?

The idea behind the absolute time reference in an iteration (snapshot) is to provide a base reference, for this iteration. Individual records in e.g. a particle can then further add offsets in time (in future versions of the standard), e.g. if one simulated in a boosted frame.

It looks to me that the use case you describe simply does not even want to store a time reference for each position, which makes indeed the base references of time confusing and wrong. Hm, maybe making them optional can indeed express that a time evolution is unknown if you really decide not to add it as a particle record.

A related issue is #141.

DavidSagan:
> Can you elaborate on "particle positions are recorded at constant longitudinal position". Wouldn't these positions be stored over a iteration (snapshot) that can be referred to a time again? You mean e.g. a track of a particle within a single iteration (snapshot) group?

Many accelerator physics codes track using the longitudinal position (often denoted "s") as the independent coordinate instead of time. Thus typically one would record particle positions for the coordinates of the particles as they reach some given s-position (typically the edge of an element).

> The idea behind the absolute time reference in an iteration (snapshot) is to provide a base reference, for this iteration. 

If, for example, the positions of multiple bunches are recorded each bunch will have its own time reference so there would not be an overall time reference for the snapshot.

> Individual records in e.g. a particle can then further add offsets in time (in future versions of the standard), e.g. if one simulated in a boosted frame.

This will be needed by the `Accelerator` extension so if there is no provision in the base standard then, for now, I will make sure it gets in the `Accelerator` extension.

ax3l:
That sounds like a plan.
In case we get #141 implemented in 2.0 this might be able to be expressed in the base standard and we might not need this attributes at all (or at a different location).

DavidSagan:
@ax3l 

>In case we get #141 implemented in 2.0 this might be able to be expressed in the base standard and we might not need this attributes at all (or at a different location).

[...]

PR #201, opened 2018-11-01
Make "time", "dt", "timeOffset" and "timeUnitSI" optional

RemiLehe:
This makes the attributes "time", "dt", "timeOffset" and "timeUnitSI" optional.

*Implements issue:* #161

## Description

*As detailed description as possible.*

*What is made required, recommended, optional?*
*What concept stands behind this change?*

*Please also add an example.*

## Affected Components

- `base`

## Logic Changes

@RemiLehe please describe how a reader should handle time now. What if only few of the attributes are present? What if a mesh is read that has a *temporal* axis? #193 Or a particle record with `unitDimension` time? What are defaults in such a case and what quantities are additive?

## Writer Changes

The above attributes are now optional

- [ ] `openPMD-api`: https://github.com/openPMD/openPMD-api/...

## Reader Changes

Readers cannot explicitly assume that these attributes are present.

- [ ] `openPMD-validator`: https://github.com/openPMD/openPMD-validator/...
- [ ] `openPMD-viewer`: https://github.com/openPMD/openPMD-viewer/...
- [ ] `XDMF` (upcoming): https://github.com/openPMD/openPMD-converter/...

## Data Updater

Not needed.

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -233,9 +233,10 @@ Required Attributes for the `basePath`
 --------------------------------------
 
 In addition to holding information about the iteration, each series of
-files  (`fileBased`) or series of groups (`groupBased`) should have
-attributes that describe the current time and the last
-time step.
+files  (`fileBased`) or series of groups (`groupBased`) can have
+attributes that describe the current time and the last time step. (In the
+case where each iteration corresponds to a different time.) Thus, the 
+following attributes are *optional*:
 
  - `time`
    - type: *(floatX)*
@@ -661,6 +662,7 @@ Reminder: for scalar records the `record` itself is also the `component`.
 
   - `timeOffset`
     - type: *(floatX)*
+    - scope: *optional* (only used if `time` and `dt` are used)
     - description: the offset between the time at which this record is
                    defined and the `time` attribute of the `basePath` level.
                    This should be written in the same unit system as `time`


STANDARD.md
```diff
@@ -233,9 +233,10 @@ Required Attributes for the `basePath`
 --------------------------------------
```
ax3l, line 233: The title in line 232 still reads "Required" :)
Do we want to make this optional or recommended?

STANDARD.md
```diff
@@ -661,6 +662,7 @@ Reminder: for scalar records the `record` itself is also the `component`.
 
   - `timeOffset`
     - type: *(floatX)*
+    - scope: *optional* (only used if `time` and `dt` are used)
```
ax3l, line 665: Maybe let's move this from *Required for each `Record`* to a section optional (or *recommended* if we decide so).

STANDARD.md
```diff
@@ -233,9 +233,10 @@ Required Attributes for the `basePath`
...
-files  (`fileBased`) or series of groups (`groupBased`) should have
-attributes that describe the current time and the last
-time step.
+files  (`fileBased`) or series of groups (`groupBased`) can have
+attributes that describe the current time and the last time step. (In the
+case where each iteration corresponds to a different time.) Thus, the 
+following attributes are *optional*:
 
  - `time`
    - type: *(floatX)*
```
ax3l, line 242: let's fix the typo in line 264 at the same time: `conversion factor`

ax3l: `timeUnitSI` should be made `optional` as well but *required* if `time` or `dt` is set

Issue #212, opened 2019-09-23
Extension: Molecular Dynamics


ax3l:
@ejcjason is working on a new extension for MD-codes with a first draft [here](https://github.com/ejcjason/MDDomainExtension).

If you like, feel free to open a pull-request to the `upcoming-2.0.0` branch with it when you feel ready for review. We are likely to be able to include it in the 2.0.0 release cycle still :)

JunCEEE:
Sorry for this late reply. Thanks,  I would like to do that. I plan to open the pull-request next week.

ax3l:
Great, yes feel free to open a PR to this repo's `upcoming-2.0.0` branch when you are ready :+1:

JunCEEE:
> Great, yes feel free to open a PR to this repo's `upcoming-2.0.0` branch when you are ready 👍

Hi @ax3l , the extension pull request has been submitted:  https://github.com/openPMD/openPMD-standard/pull/218

PR #218, opened 2020-01-26
MD simulation extension

JunCEEE:
*Implements issue:* #212  @CFGrote 

## Description

- Molecular Dynamics simulation extension is introduced for MD data IO. 

*Example*
- The example data structure can be found [HERE](https://github.com/ejcjason/MDDomainExtension).

## Affected Components

- `EXT-MD`

## Writer Changes

- The writer needs to be able to write `observables` as a group under each `Iteration`, in the same level of `particles` to describe the thermodynamic parameters of the whole atomic system.

- Not yet.

## Reader Changes

- The reader needs to be able to write `observables` as a group under each `Iteration`, in the same level of `particles` to describe the thermodynamic parameters of the whole atomic system.
Example data: https://github.com/ejcjason/MDDomainExtension/blob/master/dataMD.h5

### DIFF
--- a/EXT_MD.md
+++ b/EXT_MD.md
@@ -1,0 +1,132 @@
+Domain-Specific Naming Conventions for Molecular Dynamics Simulation Codes
+=======================================================================
+
+openPMD extension name: `MD`
+
+VERSION: 0.0.3 (January 26th, 2020)
+
+Introduction
+------------
+
+This extension is specifically designed for the molecular dynamics simulation codes.
+
+The current version of this extension is suitable to allow the output of
+arbitrary simulation codes to be post-processed and compared with common
+tools and frameworks. Future versions will define a common set of required
+records and further attributes.
+
+The example data structure can be found [HERE](https://github.com/ejcjason/MDDomainExtension).
+
+Root Group
+---------------------------
+
+### Additional Attributes for the `Root` group
+
+The following additional attributes are defined in this extension.
+The individual requirement is given in `scope`.
+
+  - `forceField`
+    - type: 1-dimensional array of N (string) elements, where N is the number of force fields (interatomic potentials) implemented in the simulation.
+    - scope: *optional*
+    - description: the methods implemented in the simulation to describe the force fields (interatomic potentials). See [Interatomic Potentials Repository](https://www.ctcms.nist.gov/potentials/).
+    - example values:
+      - `eam/alloy`
+      - `lj/cut 3.0`
+      - ...
+  - `forceFieldParameters`
+    - type: 1-dimensional array of N (string) elements, where N is the number of force fields (interatomic potentials) implemented in the simulation.
+    - scope: *optional*
+    - description: the parameters specification for the `forceField` methods. See [Interatomic Potentials Repository](https://www.ctcms.nist.gov/potentials/).
+    - example values:
+      - `pair_coeff * * 1 1`
+      - `pair_coeff 1 1 Cu_mishin1.eam.alloy Cu`
+      - ...
+
+Observable Records
+---------------------------
+
+`observables` is an *optional* group that contains physical observables that are derived from the system state, i.e., thermodynamic information (e.g. temperature, energy, pressure).
+
+### Naming conventions
+
+The naming conventions conforming the naming conventions for [Scalar, Vector and Tensor Records](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md#naming-conventions) in [openPMD base standard](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md).
+
+- `scalar` record
+  - type: *(any type)*
+  - data set: `recordName` unique name in group `basePath` + `observables`
+  - examples:
+    - /data/observables/temperature
+    - /data/observables/pressure
+
+### Attributes for each `observable` record
+
+The attributes of unit system for records should be included as defined in [openPMD base standard](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md#unit-systems-and-dimensionality).
+
+
+Particle Records
+---------------------------
+
+### Additional attributes for each `position` record
+
+The following attributes are defined in this extension. The individual requirement is given in `scope`. 
+
+- `coordinate`
+    - type: *(string)*
+    - scope: *required*
+    - description: the representation of the stored position coordinates
+    - available values:
+      - `absolute` the unscaled coordinates
+      - `fractional` the coordinates that are scaled in the range of [0,1] relative to the length of each box edge; in this coordinate system, the `unitSI` of each position component should be `1.0`
+
+### Additional Sub-Group for each Particle Species
+
+`box` is an *optional* sub-group for each particle species to contain the information of the simulation box. 
+
+### Attributes for Sub-Group `box`
+
+The following attributes are defined in this extension.
+If the `box` sub-group exists, the following attributes within it are required.
+The individual requirement is given in `scope`. 
+
+  - `dimension`
+    - type: *(uint32)*
+    - scope: *required*, if `box` sub-group exists
+    - description: the spacial dimension **D** of the simulation box.
+    - example values:
+      - `2` 2D simulation box
+      - `3` 3D simulation box
+      - ...
+
+  - `boundary`
+    - type: array of *(string)* containing **D** elements, where **D** is the value of `dimension`.
+    - scope: *required*, if `box` sub-group exists
+    - description: the boundary condition of the box along each dimension. The allowed values in `boundary` are either **periodic** or **none**.
+    - example values:
+      - `["periodic","periodic","periodic"]`  periodic in all the three dimensions
+      - `["none","periodic","periodic"]` periodic in only *y* and *z* dimensions
+
+### Records for Sub-Group `box`
+
+The following records are defined in this extension.
+If the `box` sub-group exists, the following records within it are required.
+The individual requirement is given in `scope`. 
+  - `edge`
+    - type: DxD array of *(floatX)*, where **D** is the value of `dimension`.
+    - scope: *required*, if `box` sub-group exists
+    - description: the edge vector of the simulation box in each dimension.
+    - example values:
+      - `[[1.,0.,0.],[0.,1.,0.],[0.,0.,1.]]` 3D cubic simulation box, Ax = (1.,0.,0.), Ay = (0.,1.,0.), Az = (0.,0.,1.)
+      - `[[3.46,0.,0.],[1.73,2.997,0.],[0.,0.,10.]]` 3D triclinic simulation box, Ax = (3.46,0.,0.), Ay = (1.73,2.997,0.), Az = (0.,0.,10.)
+      - `[[1.,0.],[1.,1.]]` 2D rectangle simulation box, Ax = (1.,0.), Ay = (0.,1.)
+  - `limit`
+    - type: Dx2 array of *(floatX)*, where **D** is the value of `dimension`.
+    - scope: *required*, if `box` sub-group exists
+    - description: the starting and the ending of each edge vector.
+    - example values:
+      - `[[0.,300.],[0.,150.],[0.,180.]]`  A 3D box example: xlo = 0, xhi = 300, ylo = 0, yhi = 150, zlo = 0, zhi = 180 
+      - `[[0.,300.],[15.,280.]]` A 2D box example: xlo = 0, xhi = 300, ylo = 15, yhi = 280
+  - `unitSI`
+    - type: *(float64)*
+    - scope: *required*, if `box` sub-group exists
+    - description: unit-conversion factor to convert simulation unit to SI units
+    - example: `1.0e-10`
--- a/EXT_MD.md
+++ b/EXT_MD.md
@@ -1,5 +1,5 @@
 Domain-Specific Naming Conventions for Molecular Dynamics Simulation Codes
-=======================================================================
+==========================================================================
 
 openPMD extension name: `MD`
 
@@ -8,7 +8,7 @@ VERSION: 0.0.3 (January 26th, 2020)
 Introduction
 ------------
 
-This extension is specifically designed for the molecular dynamics simulation codes.
+This extension is specifically designed for molecular dynamics simulation codes.
 
 The current version of this extension is suitable to allow the output of
 arbitrary simulation codes to be post-processed and compared with common
@@ -18,9 +18,9 @@ records and further attributes.
 The example data structure can be found [HERE](https://github.com/ejcjason/MDDomainExtension).
 
 Root Group
----------------------------
+----------
 
-### Additional Attributes for the `Root` group
+### Additional Attributes for the *Root* (`/`) Group
 
 The following additional attributes are defined in this extension.
 The individual requirement is given in `scope`.
@@ -43,13 +43,13 @@ The individual requirement is given in `scope`.
       - ...
 
 Observable Records
----------------------------
+------------------
 
 `observables` is an *optional* group that contains physical observables that are derived from the system state, i.e., thermodynamic information (e.g. temperature, energy, pressure).
 
 ### Naming conventions
 
-The naming conventions conforming the naming conventions for [Scalar, Vector and Tensor Records](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md#naming-conventions) in [openPMD base standard](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md).
+The naming conventions conforming the naming conventions for [Scalar, Vector and Tensor Records](STANDARD.md#naming-conventions) in [openPMD base standard](STANDARD.md).
 
 - `scalar` record
   - type: *(any type)*
@@ -60,11 +60,11 @@ The naming conventions conforming the naming conventions for [Scalar, Vector and
 
 ### Attributes for each `observable` record
 
-The attributes of unit system for records should be included as defined in [openPMD base standard](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md#unit-systems-and-dimensionality).
+The attributes of unit system for records should be included as defined in [openPMD base standard](STANDARD.md#unit-systems-and-dimensionality).
 
 
 Particle Records
----------------------------
+----------------
 
 ### Additional attributes for each `position` record
 
--- a/EXT_MD.md
+++ b/EXT_MD.md
@@ -70,10 +70,10 @@ Particle Records
 
 The following attributes are defined in this extension. The individual requirement is given in `scope`. 
 
-- `coordinate`
+- `positionFormat`
     - type: *(string)*
     - scope: *required*
-    - description: the representation of the stored position coordinates
+    - description: the format of the stored position coordinates
     - available values:
       - `absolute` the unscaled coordinates
       - `fractional` the coordinates that are scaled in the range of [0,1] relative to the length of each box edge; in this coordinate system, the `unitSI` of each position component should be `1.0`
@@ -118,11 +118,11 @@ The individual requirement is given in `scope`.
   - `edge`
     - type: DxD array of *(floatX)*, where **D** is the value of `dimension`.
     - scope: *required*, if `box` sub-group exists
-    - description: the edge vector of the simulation box in each dimension.
+    - description: the edge direction vector of the simulation box in each dimension.
     - example values:
-      - `[[1.,0.,0.],[0.,1.,0.],[0.,0.,1.]]` 3D cubic simulation box, Ax = (1.,0.,0.), Ay = (0.,1.,0.), Az = (0.,0.,1.)
+      - `[[1.,0.,0.],[0.,1.,0.],[0.,0.,1.]]` 3D orthorhombic simulation box, Ax = (1.,0.,0.), Ay = (0.,1.,0.), Az = (0.,0.,1.)
       - `[[3.46,0.,0.],[1.73,2.997,0.],[0.,0.,10.]]` 3D triclinic simulation box, Ax = (3.46,0.,0.), Ay = (1.73,2.997,0.), Az = (0.,0.,10.)
-      - `[[1.,0.],[1.,1.]]` 2D rectangle simulation box, Ax = (1.,0.), Ay = (0.,1.)
+      - `[[1.,0.],[0.,1.]]` 2D rectangle simulation box, Ax = (1.,0.), Ay = (0.,1.)
   - `limit`
     - type: Dx2 array of *(floatX)*, where **D** is the value of `dimension`.
     - scope: *required*, if `box` sub-group exists
--- a/EXT_MD.md
+++ b/EXT_MD.md
@@ -100,10 +100,15 @@ The individual requirement is given in `scope`.
   - `boundary`
     - type: array of *(string)* containing **D** elements, where **D** is the value of `dimension`.
     - scope: *required*, if `box` sub-group exists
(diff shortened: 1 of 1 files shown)


ax3l:
Thank you for the PR!
Before we start the review, can you please rebase against the `upcoming-2.0.0` branch and squash your commits into one?

so basically you do:
```
$ git remote -v
# ... list of remotes, remember the name of https://github.com/openPMD/openPMD-standard

$ git rebase -i <nameFromAbove>/upcoming-2.0.0
# now a text edit opens.
#   replace the first word "pick" in each line with "squash"
#   save and close the editor
# a new editor window opens with all commit messages
#   edit and keep the one you want to keep, save

$ git push -f
```
In case anything goes south in the process: first create a backup branch and you can always `git rebase --abort` as it runs.

I'll be on a working trip next week but try to take a look as soon as the PR is updated.

JunCEEE:
@ax3l Thanks. I've followed your instruction and squashed the commits. Please check that.

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
+Domain-Specific Naming Conventions for Molecular Dynamics Simulation Codes
+=======================================================================
```
ax3l:
```suggestion
Domain-Specific Naming Conventions for Molecular Dynamics Simulation Codes
==========================================================================
```

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+
+openPMD extension name: `MD`
+
+VERSION: 0.0.3 (January 26th, 2020)
+
+Introduction
+------------
+
+This extension is specifically designed for the molecular dynamics simulation codes.
```
ax3l:
```suggestion
This extension is specifically designed for molecular dynamics simulation codes.
```

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+
+The current version of this extension is suitable to allow the output of
+arbitrary simulation codes to be post-processed and compared with common
+tools and frameworks. Future versions will define a common set of required
+records and further attributes.
+
+The example data structure can be found [HERE](https://github.com/ejcjason/MDDomainExtension).
+
+Root Group
+---------------------------
```
ax3l:
```suggestion
Root Group
----------
```

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Additional Attributes for the `Root` group
```
ax3l:
```suggestion
### Additional Attributes for the *Root* (`/`) Group
```

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+
+The following additional attributes are defined in this extension.
+The individual requirement is given in `scope`.
+
+  - `forceField`
```
ax3l, line 28: for `forceField` and `forceFieldParameters`, do you want to define the syntax of the values further or keep values as a human-readable-only free text?
You can check out the [`SpeciesType`](https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/EXT_SpeciesType.md) extension for ideas how to define syntax and alternatives.

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+    - type: 1-dimensional array of N (string) elements, where N is the number of force fields (interatomic potentials) implemented in the simulation.
+    - scope: *optional*
+    - description: the parameters specification for the `forceField` methods. See [Interatomic Potentials Repository](https://www.ctcms.nist.gov/potentials/).
+    - example values:
+      - `pair_coeff * * 1 1`
+      - `pair_coeff 1 1 Cu_mishin1.eam.alloy Cu`
+      - ...
+
+Observable Records
+---------------------------
```
ax3l:
```suggestion
Observable Records
------------------
```

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+
+`observables` is an *optional* group that contains physical observables that are derived from the system state, i.e., thermodynamic information (e.g. temperature, energy, pressure).
```
ax3l, line 48: Just to clarify: Where is this group located? in the *root* (`/`) path as well? Or the `basePath`? Or in `meshesPath`/`particlesPath`?

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Naming conventions
+
+The naming conventions conforming the naming conventions for [Scalar, Vector and Tensor Records](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md#naming-conventions) in [openPMD base standard](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md).
```
ax3l: Relative links:
```suggestion
The naming conventions conforming the naming conventions for [Scalar, Vector and Tensor Records](STANDARD.md#naming-conventions) in [openPMD base standard](STANDARD.md).
```

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+- `scalar` record
+  - type: *(any type)*
+  - data set: `recordName` unique name in group `basePath` + `observables`
+  - examples:
+    - /data/observables/temperature
+    - /data/observables/pressure
+
+### Attributes for each `observable` record
+
+The attributes of unit system for records should be included as defined in [openPMD base standard](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md#unit-systems-and-dimensionality).
```
ax3l:
```suggestion
The attributes of unit system for records should be included as defined in [openPMD base standard](STANDARD.md#unit-systems-and-dimensionality).
```

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+Particle Records
+---------------------------
```
ax3l:
```suggestion
Particle Records
----------------
```

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+
+The following attributes are defined in this extension. The individual requirement is given in `scope`. 
+
+- `coordinate`
+    - type: *(string)*
+    - scope: *required*
+    - description: the representation of the stored position coordinates
+    - available values:
+      - `absolute` the unscaled coordinates
+      - `fractional` the coordinates that are scaled in the range of [0,1] relative to the length of each box edge; in this coordinate system, the `unitSI` of each position component should be `1.0`
```
ax3l, line 79: For fractional, do you intentionally include the upper value? (E.g. `[0.0, 1.0]` vs. `[0.0, 1.0)`) Just double-checking.
Formatting suggestion:
```suggestion
      - `fractional` the coordinates that are scaled in the range of `[0.0, 1.0]` relative to the length of each box edge; in this coordinate system, the `unitSI` of each position component should be `1.0`
```

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+    - type: array of *(string)* containing **D** elements, where **D** is the value of `dimension`.
+    - scope: *required*, if `box` sub-group exists
+    - description: the boundary condition of the box along each dimension. The allowed values in `boundary` are either **periodic** or **none**.
+    - example values:
+      - `["periodic","periodic","periodic"]`  periodic in all the three dimensions
+      - `["none","periodic","periodic"]` periodic in only *y* and *z* dimensions
+
+### Records for Sub-Group `box`
+
+The following records are defined in this extension.
```
ax3l, line 115: Do you want to store these small data sets as `records` or as `attributes`? Both is fine, just be aware that attributes can only be scalars or 1D-arrays, due to limitations in some file formats.

ax3l (commented):
I am sorry for the delay.

This looks wonderful, great work! I added a round of questions and formatting suggestions.

EXT_MD.md
JunCEEE: The ideal case is that there could be a well-defined syntax to describe the forceField, and to be read by MD simulation code directly. But in practice, for different forceFields, there will be different parameters. Like the [LJ potential](https://lammps.sandia.gov/doc/pair_lj.html) and  [EAM potential](https://lammps.sandia.gov/doc/pair_eam.html), the parameter syntax are different. 

Since I cannot understand all the potentials used in MD simulation, maybe I can start with defining the syntax from some potentials I understand? If it's other potential, just make it a human-readable-only free text. 

Should we also ask the opinion of some other experts about this?

EXT_MD.md
JunCEEE: I didn't think it deeply. But intuitively, for the atom located at the edge corner of the simulation box, it's fractional coordinate is [1.0,1.0,1.0]. Thus I set the range to be [0.0,1.0].

Please correct me if there are some occasions I missed.

EXT_MD.md
JunCEEE: Yes, I would like to use `records` here.

JunCEEE:
> I am sorry for the delay.
> 
> This looks wonderful, great work! I added a round of questions and formatting suggestions.

Thanks Axel. I've replied to your questions. Do we need more suggestions from experts in MD simulation?

ax3l:
@CFGrote @JunCEEE do you like to update the PR for merging? :)

EXT_MD.md
ax3l: Ok, I am just wondering about double-counting @JunCEEE - when a particle belongs to one box or another. But maybe I misunderstood and there is only one (simulation) box here.

EXT_MD.md
ax3l: @JunCEEE let's clarify this as well

EXT_MD.md
ax3l: Sure, feel free to reach out. Otherwise you can also just keep it undefined for now.

You can always add more attributes in openPMD and if you arrive at something that is worth standardizing for those attributes, then we can just specify them later.

ax3l:
VC meeting today:
- after initial LAMMPS experiments, further XMD and GROMACS workflows are being explored
- classic MD [atomic positions] & hybrid MD (DFT, ab initio) [additional particle attributes]
- we'll start with the start-to-end workflows needed for EU XFEL
  - as usual we'll, iterate in follow-up PRs if more people show interest and want to generalize it to their MD workflows, at the moment we focus on ours
- converter scripts: some outputs are already in HDF5. If existing data is large, a conversion could avoid a fully copy and instead use [Virtual Dataset (VDS) ](https://support.hdfgroup.org/HDF5/docNewFeatures/NewFeaturesVirtualDatasetDocs.html) and [external links](https://docs.h5py.org/en/stable/high/group.html#external-links) so create a light-weight openPMD file that re-interprets the original non-openPMD data

Action items:
- this PR will get polished with new XMD & GROMACS experience
- then we'll review & meet on it again

CFGrote:
hi @ax3l and @JunCEEE , is anything missing on this PR? just wondering. would be great to have it included in 2.0.0

JunCEEE:
@CFGrote I realized that the information of scattering factors and electronic charges related is missed in the standard. when I drafted the standard, I only considered the output of traditional/classical MD simulations, where the scattering process was absent. I am not sure whether we add it to this standard or have a different one. Any suggestions?

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+Particle Records
+----------------
+
+### Additional attributes for each `position` record
+
+The following attributes are defined in this extension. The individual requirement is given in `scope`. 
+
+- `coordinate`
```
zerothi: In the regular openmpd scheme coordinates are labelled `position`. Is there need to introduce another naming convention?

Wouldn't it be better to sub-inherit the `position` spec here?

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+      - `3` 3D simulation box
+      - ...
+
+  - `boundary`
+    - type: array of *(string)* containing **D** elements, where **D** is the value of `dimension`.
+    - scope: *required*, if `box` sub-group exists
+    - description: the boundary condition of the box along each dimension. The allowed values in `boundary` are either **periodic** or **none**.
+    - example values:
+      - `["periodic","periodic","periodic"]`  periodic in all the three dimensions
+      - `["none","periodic","periodic"]` periodic in only *y* and *z* dimensions
```
zerothi: Might I suggest we add `dirichlet` and `neumann` for a more complete basis?.

Also, I would refrain from using *y* and *z* as specifiers. When the `box` is a `(3, 3)` **D=3** one may have a skewed box where then the boundary refers to the box-boundaries which are not necessarily Cartesian directions. As below the triclinic box is suggested.

EXT_MD.md
```diff
@@ -0,0 +1,132 @@
...
+The individual requirement is given in `scope`. 
+  - `edge`
+    - type: DxD array of *(floatX)*, where **D** is the value of `dimension`.
+    - scope: *required*, if `box` sub-group exists
+    - description: the edge vector of the simulation box in each dimension.
+    - example values:
+      - `[[1.,0.,0.],[0.,1.,0.],[0.,0.,1.]]` 3D cubic simulation box, Ax = (1.,0.,0.), Ay = (0.,1.,0.), Az = (0.,0.,1.)
+      - `[[3.46,0.,0.],[1.73,2.997,0.],[0.,0.,10.]]` 3D triclinic simulation box, Ax = (3.46,0.,0.), Ay = (1.73,2.997,0.), Az = (0.,0.,10.)
+      - `[[1.,0.],[1.,1.]]` 2D rectangle simulation box, Ax = (1.,0.), Ay = (0.,1.)
+  - `limit`
```
zerothi, line 126: I don't understand the difference between `limit` and `edge` they are both required?
Why is `limit` needed? Could you make an example?

zerothi (commented):
General comments, I come from the DFT community and would like to explore whether OpenPMD could be of use to us!
For that this PR seems very much needed.

JunCEEE:
> General comments, I come from the DFT community and would like to explore whether OpenPMD could be of use to us! For that this PR seems very much needed.

Hello @zerothi! Welcome! Your comments would be important for us to see if this MD standard can also fulfill the need of DFT community.

zerothi:
I would be happy to review a bit more. Is there something in particular that needs attention?

EXT_MD.md
JunCEEE: Here the `coordinate` attribute is a string indicating if the `position` records are in **absolute coordinates** or **fractional coordinates**. It's not a replacement/equivalent of `position`. 

Maybe the name is a little confusing. We replace it with a bool attribute, `fractionalCoordinate`. Do you think it would be clearer to be with the bool attribute?

zerothi: If it relates directly to `position`, then why not `positionFormat`? One could also imagine scaled coordinates according to some lattice parameter.

EXT_MD.md
JunCEEE: Good suggestions. I'll add the two conditions. Also, the other dimensions will be referred to as the second and third dimensions in the example.

EXT_MD.md
JunCEEE: `edge` is a collection of the edge direction vector of each dimension of the box. It basically defines the direction of each edge/dimension of the box. `edge=[[1.,0.,0.],[0.,1.,0.],[0.,0.,1.]]` means a 3D orthorhombic (instead of cubic, my fault.) simulation box. `edge= [[3.46,0.,0.],[1.73,2.997,0.],[0.,0.,10.]]` means a 3D triclinic simulation box. 

`limit` defines the starting and ending point on each edge of the box. It basically defines the size and position of the box. 

Maybe, still, the naming is a bit confusing. It would be nice if you could suggest something better.

EXT_MD.md
JunCEEE: It sounds like a good suggestion.

JunCEEE: Is it better now?
https://github.com/JunCEEE/openPMD-standard/blob/9c769f03d097e5b87921f63ecfd0f88c71d30477/EXT_MD.md?plain=1#L73-L79

EXT_MD.md
JunCEEE: Suggestions implemented:
https://github.com/JunCEEE/openPMD-standard/blob/9c769f03d097e5b87921f63ecfd0f88c71d30477/EXT_MD.md?plain=1#L100-L111

EXT_MD.md
zerothi: I still don't understand! ;) You write exactly what the document says.
Why isn't `edge` fully explaining the `box`? What does `limit` tell you that `edge` doesn't?  
Could you make a picture (2D I guess would suffice)?

EXT_MD.md
```diff
@@ -0,0 +1,137 @@
...
+Particle Records
+----------------
+
+### Additional attributes for each `position` record
+
+The following attributes are defined in this extension. The individual requirement is given in `scope`. 
+
+- `positionFormat`
+    - type: *(string)*
+    - scope: *required*
```
zerothi, line 75: Ok, another ping here ;)

Since generally `position` is considered `absolute` wouldn't it make sense to make this optional, thus defaulting to `absolute`?

EXT_MD.md
zerothi: I agree with @JunCEEE here. It may be valuable to have an atom at the box-edge on one side and not the other. I guess the code should it-self check for mirrored positions. Especially considering that not all directions need to be periodic `[0 ; 1]` seems like the correct choice here.

EXT_MD.md
zerothi: I.e. what is the relation between `edge` and `limit`?

JunCEEE: Thanks for asking. Actually, I just realized the problem of this definition. 

The idea was inspired by `xhi` and `xlo` in https://docs.lammps.org/create_box.html. However, I didn't really implement the iead clearly in this definition. What I really need (as shown in the graph) is actually `edge` for length and direction and `origin` instead of `limit` for the origin of the box in a cartesian coordinate. Please let me know how you think if we implement it in this way.
image

zerothi: I think this is not the appropriate way. I think the lammps input is governed by the limitation of how their boxes may be defined.

Effectively your options can be superseeded by an `origin` and a `box`/`cell`/`lattice`/`vectors` keyword describing your so-called `edge` vectors (I don't think edge should be used, it is too vague I think.

I think the `vectors` seems like the best option since it lives in the `box` keyword. The `vectors` keyword then contains a `(D, D)` list of vectors with `(i, D)` being the ith vector. The entire box starts at `origin`.

EXT_MD.md
JunCEEE: It would be a good idea for a program that writes in this format. However, as a data format, I feel it's better to make the `positionFormat` clear in metadata to ensure the data is self-described. Thus I tend to keep it *required*.

What do you think?

EXT_MD.md
JunCEEE: Agree. I think we should replace `edge` with `vectors`. 

In addition to that, I'd also introduce a `lengths` keyword to define the length of box edges to ease the defining and changing of box sizes (`vectors` will only represent the direction.). Considering a box with `vectors = [[3.46,0.,0.],[1.73,2.997,0.],[0.,0.,10.]]`, it's not straightforward to change the box size to 5x5x5 using only vectors representing both directions and lengths. With the `lengths` keyword, this will be more straightforward and practical.

It will be like this:
https://github.com/JunCEEE/openPMD-standard/blob/85369d1dd4119bd27413de90a64caccaac604520/EXT_MD.md?plain=1#L125-L139

What do you think?

zerothi: I don't think `length` should be there at all. The `vectors` should contain the length of the vectors.
I don't understand your:

> Considering a box with vectors = [[3.46,0.,0.],[1.73,2.997,0.],[0.,0.,10.]], it's not straightforward to change the box size to 5x5x5 using only vectors representing both directions and lengths. With the lengths keyword, this will be more straightforward and practical.

I really have no clue what you mean here? Do you just want each vector to have a length of 5?

One parameter is more than enough to cover all cases.

My problem with your proposal is that it requires users to do this: 1) read vectors, 2) normalize vectors, 3) read lengths, 4) scale vectors
With only 1 parameter you just read it.

EXT_MD.md
zerothi: I don't know what the OpenPMD community has of *rules* for these somewhat optional things. I don't mind having it required, it is just weird given that the OpenPMD specification without the MD simulation extension does not have this. If anything this format shouldn't belong to the MD extension.

EXT_MD.md
zerothi: Actually, I don't particularly like this constraint in the name. ;)

Consider a case where one wishes to find the *ground state* of an atomic configuration. This is not a `forceField` but just a `method`. Would this make sense to change?

EXT_MD.md
JunCEEE: Thanks. `lengths` is intended to help users change the box size easily: only change the `lengths` without re-calculate the vectors, and sizes are changed more frequently than vector directions. But surely your comment also makes sense. The implementation about being user-friendly can also be done on the user interface level. If we follow the principle of storing essential information, we should set only one parameter.

zerothi: Ok. I think this should be a data format. Not a convenience format for simulators. :) So keeping things simple is much more important to me ;)

EXT_MD.md
```diff
@@ -0,0 +1,137 @@
...
+    - description: the format of the stored position coordinates
+    - available values:
+      - `absolute` the unscaled coordinates
+      - `fractional` the coordinates that are scaled in the range of [0,1] relative to the length of each box edge; in this coordinate system, the `unitSI` of each position component should be `1.0`
+
+### Additional Sub-Group for each Particle Species
+
+`box` is an *optional* sub-group for each particle species to contain the information of the simulation box.
```
zerothi, line 83: Why is the box optional for each particle species? I mean the box is *globally* defined for the entire simulation box. Not by each particle specie?

EXT_MD.md
JunCEEE: I don't know if I understand correctly, but there is indeed the `scope` with `optional` in the "[PIC extension](https://github.com/openPMD/openPMD-standard/blob/9c769f03d097e5b87921f63ecfd0f88c71d30477/EXT_ED-PIC.md)" which I take as a template. And there are actually some records belonging to the topic of MD simulation but not essential (e.g. `observables`).

Maybe @ax3l can comment on this?

ax3l: Thanks for thinking about this!

What I understand is that this attribute tries to modify how we interpret position records. While we generally can do such modifications in extensions, they might be a bit tricky to keep neatly compatible with the base standard.

For the specific task here, it looks to me like you try to achieve a "fine" and "coarse" position, similar to what we do in some particle-in-cell codes. For this purpose, the base standard has already he "position" and "positionOffset" records defined. Could we potentially use exactly that mechanism here or does it differ?

As an example, in PIConGPU, we defined `positionOffset` to a cell index, scaled via `unitSI` back to `m`. Then we use `position` with values between `[0,1)` to store the fine-grained position in a cell. The `unitSI` for `position` can be a different value than the one for `positionOffset`.

For data sets that do not need this splitting in fine and coarse positions, we define `positionOffset` to zero values (constant record).

EXT_MD.md
ax3l: Sounds like you arrived at a solution :)

EXT_MD.md
ax3l: What do you think @JunCEEE? Sounds reasonable to me, if storing the used method is the intent for this attribute :)

Note that if we use a very general attribute name like `method`, we need to add a better description to compensate what we mean :)

ax3l:
@JunCEEE @zerothi thank you for iterating on this, that is great input!
Do you two like to update the PR together to get it closer to merge-able state? :) It looks to me like most threads started arrived at a solution/improvement suggestion now.

Issue #237, opened 2021-01-07
Extension: ParticleWeighting


ax3l:
For more general usage, we could separate out from the `ED-PIC` extension the macro-particle defines into it's own extension.

Rationale: moving it to the base standard could be a bit too complicated for simple readers.

Potential non-PIC users: @AngelFP

ax3l:
Implemented by #242

PR #242, opened 2021-02-01
EXT: ParticleWeighting

ax3l:
## Description

This moves parts of the ED-PIC extension into its own extension for reuse in other domains.
    
We already conditionally honor weightings in tools like openPMD-viewer and its not specific to PIC simulations.

*Implements issue:* #237

## Affected Components

- [x] `base`
- [x] EXT:  `ED-PIC`
- [x] EXT:  `ParticleWeighting`

## Logic Changes

The definition of the `weighting` record and the `macroWeighted` and `weightingPower` attributes is moved from ED-PIC to its own extension.
ED-PIC now depends on the new extension, so there is no change to the ED-PIC extension on a logical level.

## Writer Changes

- [x] `openPMD-api`: https://github.com/openPMD/openPMD-api: not affected, because added by users manually

## Reader Changes

- [ ] `openPMD-validator`: https://github.com/openPMD/openPMD-validator can be generalized to check weighting if either `ED-PIC` or `ParticleWeighting` (or both) are declared
- [x] `openPMD-viewer`: https://github.com/openPMD/openPMD-viewer already applies weighting conditionally if found; can be generalized to check the extension first
- [x] `yt`: https://github.com/yt-project/yt not affected
- [x] `VisIt`: https://github.com/openPMD/openPMD-visit-plugin not affected
- [x] `converter`: https://github.com/openPMD/openPMD-converter not affected
- [x] `openPMD-api`: https://github.com/openPMD/openPMD-api not affected, because read by users manually

## Data Updater

- [x] https://github.com/openPMD/openPMD-updater no update needed

### DIFF
--- a/EXT_ED-PIC.md
+++ b/EXT_ED-PIC.md
@@ -18,6 +18,14 @@ simulations) that will allow to restart the checkpoints of one PIC code with
 an other.
 
 
+Dependencies
+------------
+
+This extension depends on the [ParticleWeighting](EXT_ParticleWeighting.md) extension.
+
+It is sufficient for writers to only declare the `ED-PIC` extension to express this dependency. Thus, reading implementations need to be aware that a declared `ED-PIC` extension implies `ParticleWeighting` is also applied.
+
+
 Mesh Based Records (Fields)
 ---------------------------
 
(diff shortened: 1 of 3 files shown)


AngelFP (approved):
Hi @ax3l , looks great to me!
For codes implementing openPMD to read/write, the only change I see from this is whether do declare (or look for) this extension in the `openPMDextension` attribute, right?

ax3l:
@AngelFP exactly - so if you want to use this as a writer, you add the `ParticleWeighting` string in the semicolon-separated list of the [openPMDextension](https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/STANDARD.md#hierarchy-of-the-data-file) attribute and then declare the things written in this extension.

For readers, they check if the `openPMDextension` contains the `ParticleWeighting` string (or `ED-PIC`) and if it's present, the conventions for weighting are present.

RemiLehe (approved):

Issue #238, opened 2021-01-07
Correlate Derived Mesh with Particle Group


ax3l:
For species-derived fields (density, charge density, etc.) we don't currently correlate the name with the particle species.

VisualPIC as post-processing tool (@AngelFP) correlates the two in its gui, so it would be great to hint this explicitly.

One way to achive this: consistent naming scheme (e.g. PIConGPU & WarpX: `<species>_<filter>_<fieldName>`; FBPIC: `<fieldName>_<species>`).

Another way: define a string attribute on the field with the species name (nice because simple and non-breaking).

One could even define this in the base standard for meshes.

PR #243, opened 2021-02-01
Base: Correlate Mesh & Particle Groups

ax3l:
## Description

This allows to add an optional attribute to particle groups and mesh records to cross-correlate projections and probes.

*Implements issue:* #238

## Affected Components

- [x] `base`

## Logic Changes

TODO: (write this)

## Writer Changes

None

## Reader Changes

TODO: (write this)

- [x] `openPMD-validator`: 
- [ ] `openPMD-viewer`: 
- [ ] `VisualPIC`: https://github.com/AngelFP/VisualPIC
- [x] `yt`: none
- [x] `VisIt`: none
- [x] `converter`: none
- [x] `openPMD-api`: none

## Data Updater

None

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -457,6 +457,20 @@ The following attributes must be stored with each `scalar record` and each
                    the same dimensionality N as in `gridSpacing` and
                    `gridGlobalOffset`
 
+### Optional Attributes for each `mesh record`
+
+  - `particleList`
+    - type: *(string)*
+    - description:
+        this adds a hint to analysis tools that this mesh record is related to one or more particle groups;
+        multiple particle groups can be indicated at the same time and must appear as semicolon-separated list
+    - note:
+        this is often used in graphical user-interfaces to group projected meshes of particle distributions in a user-friendly manner;
+        examples are particle densities and phase space projections
+    - examples:
+      - `electrons`: indicates this field is related to the `particle group` named `electrons`
+      - `electrons;hydrogen;carbon`: indicates this field is related to the `particle group`s named `electrons`, `hydrogen` and `carbon`
+
 
 Particle Records
 ----------------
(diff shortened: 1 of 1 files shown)


AngelFP (commented):
Thanks for this nice addition, @ax3l . This would definitely simply some things in VisualPIC!

I have one question though: the implementation here works well to link mesh records to particle groups, but this means that both have to exist. However, it can happen sometimes that, for example, the simulation output contains the charge density field associated to a particle species, but does not contain the species itself as particle group. In this case, the mesh is therefore not linked to a particle group, but simply to a species in the simulation. Would this be a problem here?

ax3l:
That's a good point. I think we can add a note for implementers to clarify.

It's still really useful to reference a particle group name for meshes like `<species>_chargeDensity`, even if the particle output is disabled for an openPMD series.

STANDARD.md
```diff
@@ -457,6 +457,20 @@ The following attributes must be stored with each `scalar record` and each
...
+
+  - `particleList`
+    - type: *(string)*
+    - description:
+        this adds a hint to analysis tools that this mesh record is related to one or more particle groups;
+        multiple particle groups can be indicated at the same time and must appear as semicolon-separated list
+    - note:
+        this is often used in graphical user-interfaces to group projected meshes of particle distributions in a user-friendly manner;
+        examples are particle densities and phase space projections
+    - examples:
```
ax3l, line 472:
```suggestion
    - note:
        when reading this back, the particle groups listed here might not be present in some steps or are even skipped in the whole openPMD series by the writer
    - examples:
```

STANDARD.md
```diff
@@ -597,6 +611,20 @@ patch order:
...
+
+  - `meshList`
+    - type: *(string)*
+    - description:
+        this adds a hint to analysis tools that this particle group is related to one or more mesh records;
+        multiple mesh records can be indicated at the same time and must appear as semicolon-separated list
+    - note:
+        this is often used in graphical user-interfaces to group meshes that are sampled with particles in a user-friendly manner;
+        examples are particle probes as well as non-uniform and curved samples from large meshes
+    - examples:
```
ax3l, line 628:
```suggestion
    - note:
        when reading this back, the mesh records listed here might not be present in some steps or are even skipped in the whole openPMD series by the writer
    - examples:
```

AngelFP:
Looks good, thanks a lot!

> It's still really useful to reference a particle group name for meshes like `<species>_chargeDensity`, even if the particle output is disabled for an openPMD series.

That's also a nice convention. I'll keep it in mind for when I have to implement it in Wake-T.

AngelFP (approved):

RemiLehe (approved):

PR #244, opened 2021-02-04
Data-Format Specific Conventions

ax3l:
## Description

This pull-request specifies data-format specific conventions.

Depending on the data-format that is targeted, a few corner-cases need to be handled to implement additional data types or sometimes central concepts of openPMD in a compatible way.

These conventions are already established in existing implementations (openPMD-api, newer codes we supported over the last years, legacy implementations in PIConGPU, libSplash, etc.). Transitioning from tribal knowledge to proper documentation, these are now specified properly.

## Affected Components

- `base`

## Logic Changes

There are no logic changes introduced that are not already used in the community.

## Writer Changes

- We now properly document that writers for HDF5 need to be compatible with `h5py` conventions and add code examples. The convention herein are used for many years in libSplash, PIConGPU, now openPMD-api and all Python implementations (including Warp, FBPIC).
- We now properly document the successful implementation details of openPMD in JSON as provided by openPMD-api by @franzpoeschel 
- We now properly document the handling of `bool` types in ADIOS1 & 2.

## Reader Changes

There are no logic changes introduced that are not already used in community codes.

## Data Updater

None needed.

### DIFF
--- a/FORMAT_ADIOS.md
+++ b/FORMAT_ADIOS.md
@@ -1,0 +1,34 @@
+# Implementation Details: ADIOS1/2
+
+## Datatype Conventions
+
+### Boolean Types
+
+ADIOS1 (BP3) was traditionally a C library and as such as no notion for boolean types.
+[At the moment](https://github.com/ornladios/ADIOS2/issues/2606), also ADIOS2 (BP4) does not define boolean types.
+Although the openPMD-standard avoids using boolean types itself, developers might choose to implement them as attributes or dataset types.
+
+#### ADIOS1
+
+ADIOS1 is deprecated.
+
+For compatibility, boolean types shall be written as `adios_unsigned_byte`.
+
+On the read side, a stored boolean type is identified as `adios_unsigned_byte`, since no convention for bool-identification is specified.
+
+#### ADIOS2
+
+For compatibility, boolean types shall be stored as `unsigned char`.
+ADIOS2 internally identifies this as `uint8_t`.
+
+On the read side, a stored boolean *attribute* is associated with an attribute itentifying it as boolean for data readers.
+The additional attribute is prefixed with `__is_boolean__/` and then adding the full path of the described attribute.
+
+Output from `bpls -A` for a boolean attribute `pybool` stored in the location of the root group (path `/`):
+```
+  uint8_t         /pybool                      attr
+  uint8_t         __is_boolean__/pybool        attr
+```
+
+There is no convention yet for a unique representation of ADIOS2 variables with boolean type.
+Thus, implementations should cast the data to and from `unsigned char` instead.
--- a/FORMAT_JSON.md
+++ b/FORMAT_JSON.md
@@ -1,0 +1,128 @@
+# Implementation Details: JSON
+
+## General Notes
+
+JSON is a format built on attribute-value pairs (dictionaries) and arrays.
+It is most commonly represented as text.
+
+Thus, in order to differentiate between openPMD **attributes** and **data sets**, slight naming conventions need to be introduced.
+An example implementation is provided in [openPMD-api](https://github.com/openPMD/openPMD-api) and in doubt, its conventions shall apply.
+
+The following JSON keys (names) are reserved for openPMD usage with JSON:
+- `attributes`
+- `data`
+- `datatype`
+
+## Datatype Conventions
+
+An openPMD **data set** is represented by a JSON object with keys `attributes`, `data` and `datatype`:
+```
+{"attributes": <attributes>, "data": <data>, "datatype": <datatype>}
+```
+
+**attributes** are defined further below and can also appear ath the openPMD **group** level.
+
+The `data` dictionary stores values as a multi-dimensional array in [row-major order](https://en.wikipedia.org/wiki/Row-_and_column-major_order#Programming_languages_and_libraries).
+The `datatype` key-value pair is defined for identification of the string-serialized type.
+
+This is a reduced example, showing scalar openPMD record named `rho`, storing values in a 2D array:
+```json
+{
+  "data": {
+    "1": {
+      "meshes": {
+        "rho": {
+          "attributes": {
+            "axisLabels": {
+              "datatype": "VEC_STRING",
+              "value": [ "x", "y" ]
+            },
+            "geometry": {
+              "datatype": "STRING",
+              "value": "cartesian"
+            },
+            "unitDimension": {
+              "datatype": "ARR_DBL_7",
+              "value": [ 0, 0, 0, 0, 0, 0, 0 ]
+            },
+            "unitSI": {
+              "datatype": "DOUBLE",
+              "value": 1
+            }
+          },
+          "data": [
+            [ 0, 1, 2 ],
+            [ 3, 4, 5 ],
+            [ 6, 7, 8 ]
+          ],
+          "datatype": "DOUBLE"
+        }
+      },
+      "attributes": {
+        "comment": {
+          "datatype": "STRING",
+          "value": "1234"
+        }
+      }
+    }
+  }
+}
+```
+
+The shape of an openPMD record component is derived from the array inside data in combination with its `datatype`.
+
+### Complex Types
+
+Complex values are serialized as simple array `[real, imaginary]`.
+When reading complex data back, a reader needs to handle the case that the dimensionality of complex `data` is reduced by `1` to derive its `shape`.
+
+### Byte Width Lookup Table
+
+In the root of the openPMD series, an additional entry named `platform_byte_widths` shall be reserved.
+It is used to map data types to byte widths for easier identification:
+```json
+{
+  "platform_byte_widths": {
+    "BOOL": 1,
+    "CDOUBLE": 16,
+    "CFLOAT": 8,
+    "CHAR": 1,
+    "CLONG_DOUBLE": 32,
+    "DOUBLE": 8,
+    "FLOAT": 4,
+    "INT": 4,
+    "LONG": 8,
+    "LONGLONG": 8,
+    "LONG_DOUBLE": 16,
+    "SHORT": 2,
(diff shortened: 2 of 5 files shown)


FORMAT_ADIOS.md
```diff
@@ -0,0 +1,30 @@
...
+
+### ADIOS1
+
+ADIOS1 is deprecated.
+A stored boolean type is identified as `adios_unsigned_byte` on the read side, since no convention for bool-identification is specified.
+
+### ADIOS2
+
+For each boolean attribute or variable, an additionally attribute is stored adding identification for bool for data readers.
+These identifying attributes are prefixed with `__is_boolean__/` (for attributes) or `__is_boolean_var__` (for variables), and then adding the full path of the described attribute or variable.
```
ax3l, line 23: @franzpoeschel I am already defining the identifier for ADIOS2 boolean variables here as well: https://github.com/openPMD/openPMD-api/issues/921

FORMAT_ADIOS.md
```diff
@@ -0,0 +1,30 @@
...
+
+### ADIOS1
+
+ADIOS1 is deprecated.
+A stored boolean type is identified as `adios_unsigned_byte` on the read side, since no convention for bool-identification is specified.
+
+### ADIOS2
+
+For each boolean attribute or variable, an additionally attribute is stored adding identification for bool for data readers.
+These identifying attributes are prefixed with `__is_boolean__/` (for attributes) or `__is_boolean_var__/` (for variables), and then adding the full path of the described attribute or variable.
```
ax3l: @franzpoeschel I am already defining the identifier for ADIOS2 boolean variables here as well: https://github.com/openPMD/openPMD-api/issues/921

franzpoeschel: I think it was a bit unlucky on my side that I used `__openPMD_internal/useSteps` but did not put the boolean attributes in `__openPMD__internal/` as well. We could maybe try coupling this with the new attribute layout and put the boolean identifiers there once we switch to it?

FORMAT_ADIOS.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Datatype Conventions
+
+### Boolean Types
+
+ADIOS1 (BP3) was traditionally a C library and as such as no notion for boolean types.
+[At the moment](https://github.com/ornladios/ADIOS2/issues/2606), also ADIOS2 (BP4) does not define boolean types.
+Although the openPMD-standard avoids using boolean types itself, developers might choose to implement them as attributes or dataset types.
+
+For compatibility, boolean types shall be stored as `adios_unsigned_byte`/`unsigned char`.
```
franzpoeschel: ADIOS2 maps `unsigned char` to `uint8_t`. But I think that `adios_unsigned_byte` is for ADIOS1, right? Also see [here](https://github.com/ornladios/ADIOS2/blob/8dd217ec720f11cf4e93e80b2db2a6953fa7b2fe/source/adios2/common/ADIOSTypes.cpp#L183).

FORMAT_JSON.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Complex Types
+
+Complex values are serialized as simple array `[real, imaginary]`.
+When reading complex data back, a reader needs to handle the case that the dimensionality of complex `data` is reduced by `1` to derive its `shape`.
+
+### Byte Width Loopup Table
+
+In the root of the openPMD series, an additional entry named `platform_byte_widths` shall be reserved.
+It is used to mape data types to byte widths for easier identification:
```
franzpoeschel:
```suggestion
It is used to map data types to byte widths for easier identification:
```
Little typo ;)

FORMAT_JSON.md
```diff
@@ -0,0 +1,121 @@
+# Implementation Details: JSON
+
+## General Notes
+
+JSON is a format build on attribute-value pairs (dictionaries) and arrays.
+Its most common representation is as strings.
```
franzpoeschel: I don't quite understand this one. Do you mean that JSON is not a binary format, but a text file?

FORMAT_JSON.md
```diff
@@ -0,0 +1,121 @@
...
+Thus, in order to differentiate between openPMD **attributes** and **data sets** slight naming conventions need to be introduced.
+An example implementation is provided in [openPMD-api](https://github.com/openPMD/openPMD-api) and in doubt, its conventions shall apply.
+
+The following JSON keys (names) are reserved for openPMD usage with JSON:
+- `attributes`
+- `data`
+
+## Datatype Conventions
+
+The the attributes are wrapped into an `data` dictionary at the location where they would usually be stored.
```
franzpoeschel:
```suggestion
The attributes are wrapped into an `attributes` dictionary at the location where they would usually be stored.
```
`data` is for the datasets

FORMAT_JSON.md
```diff
@@ -0,0 +1,121 @@
...
+Inside the `data` dictionary, values are stored as an array.
+On the same level as the `data` dictionary, a `datatype` key-value pair is defined for identification of the string-serialized type.
```
franzpoeschel: I think this explanation is a bit convoluted? Shouldn't we rather define clearly the representation of an openPMD dataset in JSON, i.e. something like:
> An openPMD dataset is represented by a JSON object with keys `attributes`, `data` and `datatype`: `{"attributes": <attributes>, "data": <data>, "datatype": <datatype>}`. In here, `<attributes>` is [put further explanations here].
An openPMD group is represented by a JSON object with key `attributes`, and additionally for each contained dataset or subgroup a key equivalent to the dataset's/subgroup's name.

Maybe also give a note that `data` is an n-dimensional array in row-major layout.

FORMAT_ADIOS.md
```diff
@@ -0,0 +1,28 @@
+# Implementation Details: ADIOS1/2
```
franzpoeschel, line 1: I would assume, one requirement that holds for all our backends is that names of datasets and groups should not clash? Should we mention that somewhere?

FORMAT_ADIOS.md
ax3l: `uint8_t` might just be the `bpls` output since the introduction of fixes size types... let's check again @franzpoeschel

franzpoeschel: In writing: `DefineVariable<unsigned char>()` will result in a `uint8_t` variable.
In reading: `IO::VariableType()` reports `"uint8_t"` which we [convert back again](https://github.com/openPMD/openPMD-api/blob/7d3e925b87c8c238405bda20ee6e2a11f6bf5d4c/src/IO/ADIOS/ADIOS2Auxiliary.cpp#L100) to `unsigned char`.

FORMAT_ADIOS.md
ax3l: Discussed on slack: in this PR, we document the current (openPMD 1) layout and will open a new pull request once the new ADIOS2 layout is finalized (https://github.com/openPMD/openPMD-api/issues/920).

At the moment we have not yet established a convention for boolean round-trip representation for data sets (ADIOS variables) and we will either establish one: https://github.com/openPMD/openPMD-api/issues/921 or hopefully can rely on a future ADIOS2 bool type: https://github.com/ornladios/ADIOS2/issues/2606

FORMAT_JSON.md
ax3l: Ups, copy-paste mistake :)

FORMAT_JSON.md
ax3l: Good idea, thank you, clarified now.

FORMAT_ADIOS.md
ax3l: Discussed on slack: no collision risk at the moment in the base standard.

FORMAT_JSON.md
```diff
@@ -0,0 +1,128 @@
...
+
+The following JSON keys (names) are reserved for openPMD usage with JSON:
+- `attributes`
+- `data`
+- `datatype`
+
+## Datatype Conventions
+
+An openPMD **data set** is represented by a JSON object with keys `attributes`, `data` and `datatype`:
+```json
```
franzpoeschel: Maybe eliminate the `json` tag here? `{"attributes": <attributes>, "data": <data>, "datatype": <datatype>}` is no valid JSON and some formatters may not like it.

Comparison with Github's formatter:
```
{"attributes": <attributes>, "data": <data>, "datatype": <datatype>}
```

```json
{"attributes": <attributes>, "data": <data>, "datatype": <datatype>}
```

FORMAT_JSON.md
```diff
@@ -0,0 +1,128 @@
+# Implementation Details: JSON
+
+## General Notes
+
+JSON is a format build on attribute-value pairs (dictionaries) and arrays.
```
franzpoeschel:
```suggestion
JSON is a format built on attribute-value pairs (dictionaries) and arrays.
```

FORMAT_JSON.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+The shape of an openPMD record component is derived from the array inside data in combination with its `datatype`.
+
+### Complex Types
+
+Complex values are serialized as simple array `[real, imaginary]`.
+When reading complex data back, a reader needs to handle the case that the dimensionality of complex `data` is reduced by `1` to derive its `shape`.
+
+### Byte Width Loopup Table
```
franzpoeschel:
```suggestion
### Byte Width Lookup Table
```

franzpoeschel (approved):
Two little typos and one formatting thing, otherwise I'm fine with this :)

FORMAT_JSON.md
ax3l: I didn't find this concerning, yet sounds reasonable to avoid future breakage.

For JSON (and XML) we could even create [schemas](http://json-schema.org/understanding-json-schema/index.html) :)

RemiLehe (approved):

PR #230, opened 2020-10-29
Wavefront ext for 2.0.0

CFGrote:
This PR adds an extension for wavefront data that occur e.g. in coherent wavefront propagation. 

It implements issue https://github.com/PaNOSC-ViNYL/ViNYL-project/issues/16 as part of Deliverable D5.1 in the EU Horizon2020 project "Photon and Neutron Open Science Cloud" Work Package 5.

###
The standard extension is defined in EXT_WAVEFRONT.MD.

*What is introduced, removed or renamed and why?*
* Required mesh record `electricField` (complexX) describes the envelope electric field amplitude of the wavefront.
The attribute `temporal_domain` denotes the Fourier domain in which the E field is represented.  Allowed values are 'time' and 'frequency'.
In the temporal domain, the physical electric field would oscillate with the photon frequency $\omega=E_{photon}/\hbar$, where $E_{photon}$ is the photon energy which is stored in the attribute `photonEnergy`. The value of `temporal_domain` also impacts the units in which `electricField` is stored: In the temporal domain, the electric field unit is `V/m`, in the frequency domain, the unit is 
Similar, the attribute `spatialDomain` denotes the spatial Fourier domain ('r' or 'k' for Cartesian or reciprocal space, respectively.

*What is made required, recommended, optional?*
Required records:
* electricField

Required attributes:
* photonEnergy
* temporalDomain
* spatialDomain
* zCoordinate

Optional attributes:
* beamline
* radiusOfCurvatureX
* radiusOfCurvatureY
* deltaRadiusOfCurvatureX
* deltaRadiusOfCurvatureY

*What concept stands behind this change?*
The concept behind this change is to provide all data and metadata to represent
output and input data for SRW (ochubar/SRW) wavefront propagation through the WPG (samoylv/WPG) wrappers.

*Please also add an example.*

## Affected Components
None

## Logic Changes
None

## Writer Changes

*How does this change affect data writers?*
*What would a writer need to change?*
*Does this pull request change the interpretation of existing data writers?*

- `libopenPMD` (upcoming): https://github.com/ComputationalRadiationPhysics/libopenPMD/...

## Reader Changes

*How does this change affect data readers?*

*What would a reader need to change? Link implementation examples!*

- `openPMD-validator`: https://github.com/openPMD/openPMD-validator/...
- `openPMD-viewer`: https://github.com/openPMD/openPMD-viewer/...
- `yt`: https://github.com/yt-project/yt/...
[...]

PR #249, opened 2021-02-26
Ext: Wavefront

ax3l:
## Description

Rename to common naming scheme for extensions.

## Affected Components

- `base`
- `EXT-Wavefront`

## Logic Changes

The identifier of the extension, as planned for openPMD 2.0, is changed.

## Writer Changes

need to change the identifier string

## Reader Changes

need to change the identifier string

## Data Converter

not yet released and only merged since #230


CFGrote (approved):
ok for me

CFGrote (commented):
all good.

PR #252, opened 2021-03-09
Ext: PatchBasedMeshRefinement

ax3l:
Adds a simple convention for patch based mesh refinement.

## Description

This adds a naming convention to efficiently identify levels of a refined mesh.

## Affected Components

- [X] `base` (cosmetic)
- [X] `EXT: PatchBasedMeshRefinement`

## Logic Changes

Adds a naming convention through record suffixes for meshes that can be used to identify refined meshes.

Allows two ways to store a refined mesh: through one record per refinement level (e.g. for ADIOS) or through one record per refinement level and patch (for non-contiguous refinements).

## Writer Changes

A writer can suffix the meshes to express levels.

- [ ] `openPMD-api`: https://github.com/openPMD/openPMD-api
- [x] `WarpX`: https://github.com/ECP-WarpX/WarpX/pull/1842

## Reader Changes

A reader can group mesh-refined records visually or build an MR structure.

Good candidates:
- [ ] `yt`: https://github.com/yt-project/yt
- [ ] `ParaView`: ...

Unaffected for now:
- [ ] `openPMD-validator`: https://github.com/openPMD/openPMD-validator
- [ ] `openPMD-viewer`: https://github.com/openPMD/openPMD-viewer
- [ ] `converter`: https://github.com/openPMD/openPMD-converter
- [ ] `openPMD-api`: https://github.com/openPMD/openPMD-api

## Data Updater

New functionality. No data update needed.

### DIFF
--- a/EXT_MeshRefinement.md
+++ b/EXT_MeshRefinement.md
@@ -1,0 +1,36 @@
+Conventions for Mesh Refinement
+===============================
+
+openPMD extension name: `MeshRefinement`
+
+Introduction
+------------
+
+This extension defines naming conventions to identify refined meshes.
+
+Mesh Records
+------------
+
+Mesh refinement levels of a mesh record describe a higher resolution subselection of a mesh record.
+
+### Additional Naming Conventions or Each Mesh `Record`
+
+The coarsest level of a record is implicitly assigned the level `0` of a simulation.
+The record names of finer levels are suffixed with `_lvl<N>` where `<N>` is the integer of the refined level.
+
+A patch of a refinemnet level shall be a spatially hyperrectangular subselection of the previous level.
+Multiple patches might exist in a refined level.
+
+If the implemented file format supports sparse data sets, i.e. through efficient chunking of patches, the refined level must over the previous level in extend and store multiple patches through its chunking mechanism.
+
+File formats that do not support efficient storage a sparesly populated refinement level can store continguous patches on the same level with an additional suffix `_<P>` where `<P>` is the number of the (hyperrectangular) patch in the refinement level.
+
+### Additional Attributes
+
+A mesh record describing a refined level shall add the following attribute:
+
+  - `refinementRatio`
+    - type: 1-dimensional array containing N *(int)*
+            elements, where N is the number of dimensions in the mesh
+    - description: the refinement ratio compared to the prior level
+    - advice to implementors: the order of the N values must be identical to the axes in `axisLabels`
--- a/EXT_MeshRefinement.md
+++ b/EXT_MeshRefinement.md
@@ -1,12 +1,12 @@
-Conventions for Mesh Refinement
-===============================
+Conventions for Patch Based Mesh Refinement
+===========================================
 
-openPMD extension name: `MeshRefinement`
+openPMD extension name: `PatchBasedMeshRefinement`
 
 Introduction
 ------------
 
-This extension defines naming conventions to identify refined meshes.
+This extension defines naming conventions to identify meshes that are refined using nested patches.
 
 Mesh Records
 ------------
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -370,7 +370,7 @@ meshes):
 
   - `axisLabels`
     - type: 1-dimensional array containing N *(string)*
-            elements, where N is the number of dimensions in the simulation
+            elements, where N is the number of dimensions in the mesh
     - description: this attribute assigns human-readible labels for the
                    indices `i`, `j`, `k`, etc. denoting the axes of a mesh
                    `A_{i,j,k}`
@@ -392,7 +392,7 @@ meshes):
 
   - `gridSpacing`
     - type: 1-dimensional array containing N *(floatX)*
-            elements, where N is the number of dimensions in the simulation
+            elements, where N is the number of dimensions in the mesh
     - description: spacing of the grid points along each dimension (in the
                    units of the simulation); this refers to the spacing of the
                    actual record that is written to the file, not that of the
@@ -403,7 +403,7 @@ meshes):
 
   - `gridGlobalOffset`
     - type: 1-dimensional array containing N *(float64 / REAL8)*
-            elements, where N is the number of dimensions in the simulation
+            elements, where N is the number of dimensions in the mesh
     - description: start of the current domain of the simulation (position of
                    the beginning of the first cell) in simulation units
     - advice to implementors: the order of the N values must be identical to
@@ -416,7 +416,7 @@ on Unit Systems and Dimensionality, further below):
 
   - `gridUnitSI`
     - type: 1-dimensional array containing N *(float64 / REAL8)*
-            elements, where N is the number of dimensions in the simulation.
+            elements, where N is the number of dimensions in the mesh.
             The order of the N values must be identical to the axes in `axisLabels`.
     - description: unit-conversion factor to multiply each value in
                    `gridSpacing` and `gridGlobalOffset`, in order to convert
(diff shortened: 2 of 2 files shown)


franzpoeschel:
I see two major issues with the current state of mesh refinement:

1. The use of suffixes `_lvl<N>` and `_<P>` seems inconsistent in an otherwise hierarchical data standard, as it reimplements hierarchies at the naming level. Should we consider developing this in conjunction with https://github.com/openPMD/openPMD-standard/issues/115 (custom hierarchies)?
2. The feature description in the standard should be more explicit on the required capabilities of the storage formats and maybe even reconsider them. Without mesh refinement, openPMD requires (roughly) a storage format that supports attributes, hierarchies and n-dimensional datasets. With mesh refinement, the backend is additionally required to allow querying written regions (in the API, the information returned by `.available_chunks()` is purely backend-specific and completely transparent to the actual openPMD standard).
In my understanding, HDF5 chunking allows storing a sparsely written dataset without using disk space corresponding with the entire domain (rather, it will use disk space corresponding with the covered chunks). Explicitly encoding the written chunks in either the standard or the HDF5 schema would then enable mesh refinement in HDF5, too.
To me, this seems like too small a detail for making mesh-refinement ADIOS-only over it.

ax3l:
Potential directions for HDF5 support of MR in openPMD:
- https://github.com/HDFGroup/hdf5/discussions/3257
- https://github.com/LifeboatLLC/SparseHDF5

ursg:
This is clearly motivated specifically with a patch-based refinement strategy in mind, and as such probably works very well.

However, I would like to point out that this structure would seem quite alien to other refined mesh approaches (tree-based or individuall-cell-based ones). Deliberately grouping mesh data by refinement level does not come natural to them!

Sorry about this nonproductive comment, as I do not have a suggestion for a suitable, more general approach. There certainly *is* merit in the proposal as it is specified here, but you might want to adapt a more specific name ("PatchBasedMeshRefinement"), as to not encumber future applications with different approaches.

EXT_MeshRefinement.md
```diff
@@ -0,0 +1,36 @@
+Conventions for Mesh Refinement
+===============================
+
+openPMD extension name: `MeshRefinement`
```
ax3l:
```suggestion
Conventions for Patch Based Mesh Refinement
===========================================

openPMD extension name: `PatchBasedMeshRefinement`
```

EXT_MeshRefinement.md
```diff
@@ -0,0 +1,36 @@
...
+
+Introduction
+------------
+
+This extension defines naming conventions to identify refined meshes.
```
ax3l:
```suggestion
This extension defines naming conventions to identify meshes that are refined using nested patches.
```

EXT_MeshRefinement.md
```diff
@@ -0,0 +1,36 @@
...
+------------
+
+Mesh refinement levels of a mesh record describe a higher resolution subselection of a mesh record.
+
+### Additional Naming Conventions or Each Mesh `Record`
+
+The coarsest level of a record is implicitly assigned the level `0` of a simulation.
+The record names of finer levels are suffixed with `_lvl<N>` where `<N>` is the integer of the refined level.
+
+A patch of a refinemnet level shall be a spatially hyperrectangular subselection of the previous level.
```
ax3l, line 21:
```suggestion
A patch of a refinement level shall be a spatially hyperrectangular subselection of the previous level.
```

EXT_MeshRefinement.md
```diff
@@ -0,0 +1,36 @@
...
+
+If the implemented file format supports sparse data sets, i.e. through efficient chunking of patches, the refined level must over the previous level in extend and store multiple patches through its chunking mechanism.
+
+File formats that do not support efficient storage a sparesly populated refinement level can store continguous patches on the same level with an additional suffix `_<P>` where `<P>` is the number of the (hyperrectangular) patch in the refinement level.
+
+### Additional Attributes
+
+A mesh record describing a refined level shall add the following attribute:
+
+  - `refinementRatio`
```
RemiLehe, line 32: I don't think that is needed, because we already save the cell size at each level. So this would duplicate the same info.

STANDARD.md
```diff
@@ -754,6 +754,8 @@ defined:
   see [EXT_BeamPhysics.md](EXT_BeamPhysics.md).
 - **ED-PIC**: electro-dynamic/static particle-in-cell codes,
   see [EXT_ED-PIC.md](EXT_ED-PIC.md).
+- **MeshRefinement**: mesh records refined over multiple levels of resolution,
```
ax3l, line 757:
```suggestion
- **PatchBasedMeshRefinement**: mesh records refined over multiple levels of resolution,
```

EXT_MeshRefinement.md
RemiLehe, line 21:
```suggestion
A patch of a refinement level shall be a spatially hyperrectangular subselection of the previous level.
```

BenWibking:
This is supported by `openpmd-visit-reader` (https://github.com/BenWibking/openpmd-visit-reader) for VisIt, with an extension for how Parthenon (https://github.com/parthenon-hpc-lab/parthenon/pull/1050) handles level suffixes for vector components. (Parthenon's octree AMR works fine in this representation.)

This plugin has also been tested with Quokka and WarpX outputs, which AFAIK adhere more strictly to this proposal.

EXT_MeshRefinement.md
```diff
@@ -0,0 +1,36 @@
...
+
+Mesh Records
+------------
+
+Mesh refinement levels of a mesh record describe a higher resolution subselection of a mesh record.
+
+### Additional Naming Conventions or Each Mesh `Record`
+
+The coarsest level of a record is implicitly assigned the level `0` of a simulation.
+The record names of finer levels are suffixed with `_lvl<N>` where `<N>` is the integer of the refined level.
```
pgrete, line 19: Does the "implicit" imply that level `0` cannot (or may not) carry the `_lvl` suffix?

EXT_MeshRefinement.md
```diff
@@ -0,0 +1,36 @@
...
+
+A patch of a refinemnet level shall be a spatially hyperrectangular subselection of the previous level.
+Multiple patches might exist in a refined level.
+
+If the implemented file format supports sparse data sets, i.e. through efficient chunking of patches, the refined level must over the previous level in extend and store multiple patches through its chunking mechanism.
+
+File formats that do not support efficient storage a sparesly populated refinement level can store continguous patches on the same level with an additional suffix `_<P>` where `<P>` is the number of the (hyperrectangular) patch in the refinement level.
```
pgrete, line 26: Does this need to be specified at all?

For example, if I have a mesh where half of the domain is refined and the other is not, the current wording suggests a either a global mesh with sparse data or a patched based one.
I could imagine that a simpler logic (i..e, just the `_lvlN` without considering sparse or patches) could be enough as it would already allow a code to query if data at certain location (and a certain level) exist.

I would also consider keeping the `_<P>` suffix, but optional and allowing for overlapping region (without making guarantees on the data in the overlapping regions), as it would allow to write individual patches/blocks with individual ghost zones (e.g., for debugging or calculating derived quantity on fine levels without prolongation from coarser ones).

This should then also cover/provide flexibility other MR strategies without breaking the existing usage, doesn't it?

STANDARD.md
pgrete: Following up on the comment above, are there advantages to differentiate between patch-based refinement and other types of refinement?
Or could all types be covered by a very simple/basic extension?

pgrete (commented):
Dear all,
what would be required to push this across the finish line?
There are now multiple codes that use (some variant) of this extension plus some downstream tooling is available, too.
It seems like a good time to finalize the extension.

PR #262, opened 2022-01-12
Fix: RFC 2119 link

ax3l:
The link to IETF RFC 2119 is broken.

This adds a new link and changes the spelling of keywords to lowercase.

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -6,9 +6,9 @@ VERSION: **1.1.0** (Feburary 6th, 2018)
 Conventions Throughout these Documents
 --------------------------------------
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
-"SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be
-interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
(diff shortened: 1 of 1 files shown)


PrometheusPi (approved):

s9105947:
Is there a specific reason to break with the corresponding [BCP 14](https://www.rfc-editor.org/info/bcp14), which consists of [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) (keyword definition) + the 2017 update [RFC 8174](https://datatracker.ietf.org/doc/html/rfc8174) (spelling clarification) which demand to **always** use ALLCAPS?

> Personally I totally missed that the lower-case keywords are intended to follow RFC 2119 when first skimming the standard -- allthough I suppose the allcaps might be confusing when unfamiliar.

ax3l:
Yes, all-caps is just too weird to read and thus we never did that. Thus we re-define the keywords to lowercase now, otherwise same naming as indicated.

This whole definition is just a clarification, because many non-native (and some native) English speakers do not immediately see the difference between `must` and `shall`/`should`. This RFC nicely resolves this by clearly defining the meaning and difference/equivalence of such qualifiers.

ax3l:
@s9105947 is it ok if you add your name & affiliation to https://github.com/settings please ? :)
It's a bit hard otherwise for us to keep track who we email and chat on GitHub with :slightly_smiling_face:

Issue #261, opened 2022-01-04
Species Type: Clarifications on Molecule Syntax


s9105947:
The current version of the species type extension demands for molecule syntax:
"Use standard chemical notation, e.g.: H2O."

What is "standard chemical notation" specifically?

Notable cases:

- Molecules with charge, e.g. `OH-`
- complexes, e.g. `[NiCN]+`
- organic compounds, e.g. `C2O5OH`, `CH3COOH`, maybe even `CH3-COOH`, `CH3-CH2-CH3`
- *pedantic* there are more complex formulae which are "standard" too, but can't easily be translated to text, e.g. [structural formulae](https://en.wikipedia.org/wiki/Structural_formula)
- Are parenthesis allowed? e.g. `SO2(OH)2` (admittedly non-standard)
  - Note that parenthesis need to be balanced, which increases parsing complexity: With paranthesis, this **can't be parsed by regex**.
  - Semantics: Are `(H2)O` and `H2O` considered equal?
  
This is a bottomless hole, so I'd like to suggest as a minimal version:

- ("An implementation must support at least the following notation. Additionally, more specific notation may be supported:")
- A molecule is given as a **sum formula**, e.g. semantically only the amount per atom is extracted.
- Parenthesis, brackets, (and braces) of any form are forbidden.
- If an atom occurs twice its quantities are added. ("An implementation may derive additional information from this, e.g. anorganic groups like `OOH`.")
- Only the *total charge* of a molecule can be specified by appending `"^" *DIGIT ("+" / "-")`, e.g. `HCO^-`, `SO4^2-`
  - If the digit is omitted it is assumed to be one, i.e. `HCO^1-` and `HCO^-` are equivalent.
- Note: A total quantity of a molecule can not be given, i.e. `2H^+` is **invalid**.

Note: Depending on if an isotope is also an atom maybe use "atom or isotope" repectively.  
Note: Potentially replace the charge separator character.

ax3l:
The answer for this is similar to https://github.com/openPMD/openPMD-standard/issues/260#issuecomment-1006464841:

We don't have a use case with a committed chemistry code yet and leave this open to the first one(s) that would like to adopt openPMD. I like your suggestion, but would not standardize it until we have a concrete use case/adopter.

[...]

Issue #260, opened 2022-01-04
Species Type: Suggestion for Ion Syntax


s9105947:
The species type extension in the section for atoms notes:
"Any extension using this standard can define how to specify the charge state."

Would it be possible/advised to add such a definition?  
I recon it is a somewhat common usecase, and would be justified in a standard.

I'd like to suggest the following notation:

Examples: `He^2+`, `H^+`, `Cl^-`, `O^2-`, `#3He^2+`  
Generalized: `atom "^" *DIGIT ( "+" / "-" )`  
Rationale:

- use a delimiter after the atom to specifically denote charge -- **I'm not sure that `^` is a good idea,** maybe another character is better suited (`_`, `/`, `|`, `:`?)
- allow omission of amount if it is 1
- could be used with molecules too (see below)

> I'm not aware how existing implementations solve this, maybe copy from them instead?

ax3l:
>  "Any extension using this standard can define how to specify the charge state."

Yes, this is to avoid confusion and determine which components of the standard have precedence.
For instance, the `ED-PIC` convention has a clear meaning of charge, we then only add extra meaning with the `SpeciesType` extension to clarify that something is a Hydrogen or Carbon isotope.

> Would it be possible/advised to add such a definition?
> I recon it is a somewhat common usecase, and would be justified in a standard.

Unless we have a concrete chemical use case, I would not specify this yet. This is also the reason why the "Molecule" section is intentionally sparse for future use.

The reason behind this is the following: currently, most our use cases are in laser-plasma physics, astrophysics, high-energy physics and some data analysis tasks. In all the listed domain sciences, the charge state changes during the simulation, thus we need to encode it *per particle* outside the `SpeciesType` extension. (Our codes have ionization physics and in the future maybe also re-combination physics included.)

The notions you list are chemical notations and assume the charge state cannot change in a data series.

So yep, good idea to add, but not without a use case (i.e., an actively maintained chemistry code that commits to adopt openPMD). We can keep this issue open for potential future adopters if you like :)

s9105947:
This touches "What *is* a species?", and I'd agree that the charge state of individual particles is not part of that -- i.e. it does not make sense to include the charge in a species definition.

Suggestion for standard text:

The base `speciesType` standard does not support ions.
[...]

Issue #259, opened 2022-01-04
Species Type: Clarification on Atom Syntax


s9105947:
In the species type extension:
Are temporary (three-letter UuX) to be supported by default?  
E.g. "Any one or two-letter abbreviation from the periodic table, defined by..."

> I'm not sure if the three-letter symbols where ever part of the official periodic table, though a couple I had included them.

Is the case fixed?  
E.g. "The first letter must be in upper case, and the second (should it exist) in lower case." or simply "The element name is case-sensitive."

Also regarding the language:  
Is an isotope also an atom?
I.e. can isotopes be used to construct ions and molecules?

ax3l:
Q:
> Are temporary (three-letter UuX) to be supported by default?

Current standard:
> Element namings follow the abbreviated namings of the periodic table, defined by The International Union of Pure and Applied Chemistry (IUPAC). An example would be Si for silicon.

So I would say, if IUPAC gave out that temporary name, then yes, no problem. Note that temporarily here means usually ~10 years (and is easy to translate once a new name gets assigned).

> Is the case fixed?
E.g. "The first letter must be in upper case, and the second (should it exist) in lower case." or simply "The element name is case-sensitive."

Sounds good, feel free to add this in the first paragraph via a PR.

Q:
> regarding the language:
> Is an isotope also an atom?

Isotopes are atoms, we just list them explicitly in the section title and 2nd paragraph to make sure people find it.

Q:
>  I.e. can isotopes be used to construct ions and molecules?

Yes, the 2nd paragraph shows such an example.
Current standard:
> Specifications of isotopes are denoted by a pound symbol # followed by the isotopic number followed by the chemical symbol, e.g.: #3He for Helium-3.

Note that we intentionally do not encode charged (ion) vs. neutral (atom/isotope) here.
See the third paragraph.
Current standard:
> The charge state is not encoded by the speciesType attribute. Any extension using this standard can define how to specify the charge state.

Issue #258, opened 2022-01-04
Species Type: Character Set after "other:"


s9105947:
To specify custom species, the species Type extension allows: "user are free to append a free text after a colon".
Is really **any** text allowed? (What even is *text*?)

IMO "free text" should be severely restricted, at least to "printable characters without semicolon (;)".
notable cases:

- semicolons ";" -> already is species list separator
- colons ":" -> is separator after "other" (though probably harmless)
- apostrophe (') and quotation marks (")
  - are generally annoying to escape
  - might be re-written by editors to use their UTF8-versions (“ ” ‘ ’)
- backslash \\
  - is equally annoying to escape
  - is sometimes used to write control characters, e.g. `\\n` or `\\a`
- whitespaces -> sometimes annoying to handle, but generally safe
  - in order of [difficulty|annoyingness] to handle: spaces < tabs < newlines
- control characters -> influence display
  - ASCII? (e.g. nul, form feed, carriage return, backspace, bell)
  - UTF8? (e.g. byte order mark, right to left text indicator)
- non-ascii in general
  - e.g. anything non-english (e.g. german, russian, arabic, chinese)
  - (UTF8 is endless, see also: emojis, skin-tone modifiers, single flag characters, greek question mark etc.)
- *pedantic* is a character set even specified? are obscure iso-charsets allowed?
- empty string (entire name would be "other:")

By gut instinct I'd suggest to allow only "universally" **safe** and **unambigous** characters after "other:":  
- only ASCII encoding (possibly already fixed by type "string", I have not checked)
- any number of characters from any number of these classes:
  - letters a-z (either case)
  - digits 0-9
  - one of: dash "-", plus "+", underscore "_", period ".", number sign "#", caret "^"
- excluded by design:
  - ";" b/c it is the list delimiter
  - "," b/c it could be confused for a list delimiter
  - whitespaces (space, tab, newline) b/c they tend to be difficult to handle or form special cases
  - backslash, apostrophe, quotation mark b/c they are annoying to escape

ax3l:
> To specify custom species, the species Type extension allows: "user are free to append a free text after a colon".
Is really any text allowed? (What even is text?)
> IMO "free text" should be severely restricted, at least to "printable characters without semicolon (;)".

Correct. Derived from the base standard, we always mean pure ASCII when we write text.
Yes, generally we leave `other:` unspecified for cases we have not yet thought about and that get standardized in later versions.
[...]

PR #264, opened 2022-01-27
clarify species type extension

s9105947:
This PR clarifies the species type extension, adds a full definition using [ABNF](https://datatracker.ietf.org/doc/html/rfc5234).
*Technically* it introduces breaking changes -- for details, please see below.

Issues: #256 #257 #258 #259 #260 #261

## Description

- #256 Full ABNF grammar
  - I added a full definition using ABNF in every section.
  - In the future the rule for elements might have to be updated.
  - Examples are still the main way readability is ensured -- the grammar is intended as an auxiliary tool if the textual description is not sufficient.
- #257 single species and multiple species
  - There is a clearer distinction made between a single and a list of particle species.
  - **Breaking**: `other:` is *not allowed in lists*, as to allow it the semicolon `;` would have to be forbidden from its content.
  - Separator placement is clarified: Double list separator forbidden (`item1;;item2`), trailing separator optional.
  - Clarified that a list with one particle species and a single particle species **must** be treated equally -- as they can't (necessarily) be distinguished/used the same syntax.
  - We discussed encoding (relative) quantities included in lists. This is not included.
- #258 charset after `other:`
  - Clarified that only printing ascii characters (0x21-0x7e), as well as space are allowed.
  - **Breaking**: Control characters are now forbidden. Note that newline (ascii 0xa) and horizontal tab (ascii 0x9) are thereby forbidden too.
  - Added note discouraging use of `other:` (without suffix) as well as trailing spaces -- even though they are allowed.
- #259 clarify atom syntax
  - Clarified that elements must be case sensitive.
- #260 ion syntax
  - Proposition of ion syntax has been dropped.
  - Clarified that ions are not supported by the base standard.
  - Added note discussing when charge state SHOULD NOT and when it MAY be encoded by an implementation.
- #261 molecule syntax
  - Proposition for molecule syntax has been dropped.
  - Clarified by rewriting paragraph.

In its entirety, this aims to clarify the current version of the standard.
Except for the treatment of `other:`, all unambiguous parts have been left untouched.

Valid examples:

```
electron
He
#2H
electron;H
electron;H;
Si;
other:
other:myitem
other:this looks like a list;but is actually a single particle species
other: note that this text can be very long
other: there is no \n escaping
```

Invalid examples:

| species       | invalid because                       |
|---------------|---------------------------------------|
| `other:·`     | non-ascii char `·`                    |
| `electron;;H` | empty list item/double semicolon `;;` |
| `Uux`         | not part of the periodic table        |
| `H2O`         | molecules not supported               |

## Affected Components

- `EXT-SpeciesType`

## Logic Changes

- `other:` forbidden in lists
- unprecise description for molecules removed, such implementations would be non-conformant now

## Writer Changes

- Forbid non-printing ascii in `speciesType`s
- ensure lists do not contain `other:`
- ensure that particle species lists with one item are equivalent to single particle species

Existing writers are compatible, as long as they

- do not use non-printing ascii chars for `speciesType`

## Reader Changes

Lists with one item must be considered single particle species.

As molecules and ions have been explicitly forbidden, reader implementations are encouraged to specify their additionally accepted syntax explicitly.
(Which they should do anyways).

- `openPMD-validator`: https://github.com/openPMD/openPMD-validator/
  - not implemented
  - [PR #38](https://github.com/openPMD/openPMD-validator/pull/38) is broken anyways (checking against undocumented charset, no syntax checking)
  
(I did not encounter any other mentions of `speciesType` in the related tools.)

## Data Converter

Not implemented, hence no changes necessary.

### DIFF
--- a/EXT_SpeciesType.md
+++ b/EXT_SpeciesType.md
@@ -49,17 +49,17 @@ The following additional attribute for openPMD `mesh records` and
 
 This attribute can be used with any `record` (including `mesh records`).
 
-A particle species list containing only a single item MUST be treated as a
-single species.
+A particle species list containing only a single item **must** be
+treated as a single species.
 
 > Note that `other:` can *only* be used for single species, it is
 > forbidden in lists.
 >
 > The list separator is a sole `;`. Neither spaces around it nor empty
 > list items (two semicolons following each other `;;`) are allowed.
-> However, a single trailing semicolon `;` MUST be ignored.
+> However, a single trailing semicolon `;` **must** be ignored.
 
-The following values SHOULD NOT be used by an implementation, even if
+The following values **should not** be used by an implementation, even if
 they are allowed above:
 
 - `other:`, as it does not specify any content.
@@ -161,15 +161,15 @@ element = "H" / "He" / "Li" / "Be" / "B" / "C" / "N" / "O" / "F" / "Ne" / "Na" /
 The charge state is not encoded by the `speciesType` attribute.
 
 The charge state ("ionization") is an attribute of an individual
-particle, and therefore an implementation SHOULD NOT encode a charge
-state as part of the `speciesType` in any way.
+particle, and therefore an implementation **should not** encode a
+charge state as part of the `speciesType` in any way.
 In the case that the implementation does treat the charge state
 ("ionization") as a global, unmodifiable property of the `speciesType`
-it MAY encode a charge state in an implementation-defined way.
+it **may** encode a charge state in an implementation-defined way.
 
 ### Molecules
 Molecules are not supported by the `speciesType` attribute.
 
-However, an implementation MAY encode molecules using an
+However, an implementation **may** encode molecules using an
 implementation-defined syntax. In this case typical chemical notation
-SHOULD be accepted, e.g. `H2O` for water.
+**should** be accepted, e.g. `H2O` for water.
--- a/EXT_SpeciesType.md
+++ b/EXT_SpeciesType.md
@@ -9,7 +9,7 @@ Introduction
 This convention is for standardizing the names of particle species, e.g. in
 particle physics.
 
-This document uses ABNF as speciefied by [RFC 5234][rfc5234].
+This document uses ABNF as specified by [RFC 5234][rfc5234].
 
 [rfc5234]: https://datatracker.ietf.org/doc/html/rfc5234
 
--- a/EXT_SpeciesType.md
+++ b/EXT_SpeciesType.md
@@ -3,13 +3,15 @@ Convention for Specifying Particle Species
 
 openPMD extension name: `SpeciesType`
 
-
 Introduction
 ------------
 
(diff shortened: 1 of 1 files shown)


s9105947:
Following discussion in #262 I replaced RFC 2119 terms (MUST, SHOULD NOT...) with **bold text**, but using lowercase.
I can remove the emphasis, though I'm a little reversed towards that -- I used the subtle differences between must (not) and should (not) deliberately, and kinda fear that being lost. Though diverting the style from the remaining standard **must** be avoided.

PrometheusPi (approved):

EXT_SpeciesType.md
```diff
@@ -3,13 +3,15 @@ Convention for Specifying Particle Species
...
 openPMD extension name: `SpeciesType`
 
-
 Introduction
 ------------
 
 This convention is for standardizing the names of particle species, e.g. in
 particle physics.
 
+This document uses ABNF as speciefied by [RFC 5234][rfc5234].
```
ax3l:
```suggestion
This document uses ABNF as specified by [RFC 5234][rfc5234].
```

EXT_SpeciesType.md
```diff
@@ -24,76 +26,150 @@ The following additional attribute for openPMD `mesh records` and
...
 
 We currently do not define spellings of hadrons besides the commonly used ones
-below and suggest for this version to use `other:` with namings from the
-[particle data group (PDG)](http://pdg.lbl.gov/). Other means of grouping e.g.
-jets can be used, e.g. additional attributes outside of the definition of this
-extension.
+below.
+
+For particle species not mentioned here we suggest to use `other:`
+with namings from the [particle data group (PDG)](http://pdg.lbl.gov/).
```
ax3l:
```suggestion
with namings from the [particle data group (PDG)](http://pdg.lbl.gov).
```

ax3l (commented):
Thank you already!

I marked a few typos and will do a thorough review soon!

DavidSagan (approved):
Thanks for the work done by everyone!

s9105947:
If I read the [openPMD validator](https://github.com/openPMD/openPMD-validator/blob/2.0.X/openpmd_validator/check_h5.py#L851) correctly (do I?), currently only non-empty strings consisting of `a-zA-Z0-9:;-` are accepted as `SpeciesType`. Though this is not documented in the current standard, should we put it into the current version?

> My main idea of using printing ASCII chars was to allow the broadest, "non-harmful" charset to not break compatability. But I guess that the openPMD validator is also somewhat authoritative, so presumably nothing would be broken by demanding `a-zA-Z0-9;:-`?

ax3l:
I think that makes sense, if one uses the species type extension then one should actually specify a `SpeciesType`.
(We have special keywords like `None` for some attributes, but I am not sure it makes sense here.)

ax3l (approved):

ax3l:
Sorry for the delay, in it goes! Great work!

PR #271, opened 2023-01-23
Laser extension

MaxThevenet:
This PR proposes a new extension to the standard, suitable for describing a laser envelope on a regular grid.

## Description

Although a laser pulse can be described as an electric and a magnetic field, it is sometimes much more efficient (in particular because one then need to describe the envelope only, not the laser frequency) to describe it as an envelope, and store its central frequency/wavelength as metadata. This is relevant for PIC codes relying on a laser envelope (most quasi-static PIC codes, and a number of electromagnetic PIC codes), but is not covered by the ED-PIC extension. The new EXT-LaserEnvelope is short and just targets this use case (with minimal metadata).

## Affected Components

- `STANDARD.md`
- `EXT_LaserEnvelope.md`

## Reader Changes

This PR will not change any behavior, but the new option to describe a laser envelope will be incorporated with some changes in at least `openPMD-viewer`.


ax3l:
Thanks a lot for this, Maxence!

We already reviewed this in lasy and have currently no further comments on it.

Let me know if you like to keep it open for a bit as we implement in lasy or if we shall stage it into the `upcoming-2.0.0` branch already.

EXT_LaserEnvelope.md
```diff
@@ -0,0 +1,68 @@
...
+------------
+
+This extension is specifically designed for the modeling laser beam propagation.
+
+Mesh Based Records (Fields)
+---------------------------
+
+### Naming Conventions for `mesh record`s (field records)
+
+The laser pulse is represented as a single complex field $\mathcal{E}$, describing the envelope of the laser electric field in the paraxial approximation:
```
RemiLehe:
```suggestion
The laser pulse is represented as a single complex field $\mathcal{E}$, describing the envelope of the laser electric field:
```
My understanding is that we do not need to do the paraxial approximation at this point. In fact, the propagator we will implement in `lasy` may not make this approximation.

EXT_LaserEnvelope.md
```diff
@@ -0,0 +1,68 @@
...
+
+```math
+   \begin{aligned}
+   E_x &= \operatorname{Re}\left( \mathcal{E} e^{-i\omega_0t}p_x\right)\\
+   E_y &= \operatorname{Re}\left( \mathcal{E} e^{-i\omega_0t}p_y\right)\\
+   E_z &= 0\\
+   B_x &= -c E_y\\
+   B_y &= c E_x\\
+   B_z &= 0\end{aligned}
```
RemiLehe:
```suggestion
\end{aligned}
```
Would it be okay to only specify `E_x` and `E_y` here? In principle, `E_z` could be non-zero (even in the paraxial approximation, I think).

EXT_LaserEnvelope.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Additional attributes on the `mesh record` named `electricField`
+
+On the `series` object, set the following attributes:
+
+  - `angularFrequency`
+    - type: *(floatX)*
+    - description: Angular frequency $\omega_0$ at which the laser envelope is defined.
+    - scope: *required*
+    - `unitDimension = (0., 0., -1., 0., 0., 0., 0.)` (rad/s)
```
RemiLehe:
```suggestion
    - description: Angular frequency $\omega_0$ at which the laser envelope is defined (rad/s).
    - scope: *required*
```
We actually cannot specify `unitDimension` as an attribute since `angularFrequency` is already an attribute.

EXT_LaserEnvelope.md
```diff
@@ -0,0 +1,68 @@
...
+
+When added to an output, the following naming conventions shall be used for complex electric field `mesh records`.
+
+- `laser_envelope/`
+  - type: *(complexX)*
+  - scope: *(required)*
+  - decription: Scalar field for the envelope (in V/m). See above for description.
+  - unitDimension = `(1., 1., -3., -1., 0., 0., 0.)` $(V/m = kg . m / (A . s^3))$
+
+### Additional attributes on the `mesh record` named `electricField`
```
RemiLehe:
```suggestion
### Additional attributes on the `mesh record` named `laserEnvelope`
```

EXT_LaserEnvelope.md
```diff
@@ -0,0 +1,68 @@
...
+   B_x &= -c E_y\\
+   B_y &= c E_x\\
+   B_z &= 0\end{aligned}
+```
+
+where $\operatorname{Re}$ stands for real part,  $E_x$ (resp. $E_y$) is the laser electric field in the x (resp. y) direction, $\mathcal{E}$ is the complex laser envelope described in the standard, $\omega_0 = 2\pi c/\lambda_0$ is the angular frequency defined from the laser wavelength $\lambda_0$ and $(p_x,p_y)$ is the (complex and normalized) polarization vector. The polarization state (linear, circular, elliptical) is controlled by the phase of the polarization vector. For instance, if $arg(p_x) = arg(p_y)$, the polarization is linear. If $arg(p_x) = arg(p_y) + \pi/2$, the polarization is circular.
+
+When added to an output, the following naming conventions shall be used for complex electric field `mesh records`.
+
+- `laser_envelope/`
```
RemiLehe:
```suggestion
- `laserEnvelope/`
```

EXT_LaserEnvelope.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+  - `polarization`
+    - type: *(floatX)*
+    - description: Polarization vector. This vector is normalized (so its modulus is 1) and contains all information to account for polarization states and carrier-envelope phase (CEP).
+    - scope: *required*
+    - components:
+      - `x/`
+      - `y/`
+    - `unitDimension = (0., 0., 0., 0., 0., 0., 0.)` (unitless)
```
RemiLehe:
```suggestion
```
Similarly, we cannot have components on an attribute. Please have a look at `gridSpacing` in `STANDARD.md`. Here we could create a 2-component array, with well-defined order (i.e. first element is `x`, second is `y`).

EXT_LaserEnvelope.md
```diff
@@ -0,0 +1,68 @@
...
+
+  - `propagationDirection`
+    - type: *(floatX)*
+    - scope: *optional*
+    - description: Propagation direction for the laser pulse. Default is (0,0,1), i.e. propagation along the z coordinate.
+    - components:
+      - `x/`
+      - `y/`
+      - `z/`
+    - `unitDimension = (1., 0., 0., 0., 0., 0., 0.)` (length)
```
RemiLehe:
```suggestion
```

I am not sure that this attribute is actually needed. I think it would enough to specify all the other parameters, and the propagation direction would be chosen by the user in their simulation input script, independently from this file.

RemiLehe (changes requested):
@MaxThevenet I added a few comments ; would you be able to address them?

EXT_LaserEnvelope.md
MaxThevenet: I think that's correct

EXT_LaserEnvelope.md
MaxThevenet: I think non-zero Ez would be a correction to the paraxial approximation, e.g. in https://journals.aps.org/pre/abstract/10.1103/PhysRevE.58.3719. In any case, I agree it is better not to specify that these are 0 because they do not have to be.

EXT_LaserEnvelope.md
MaxThevenet: Ah thanks, I think I took it from https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/EXT_Wavefront.md?plain=1#L46.

EXT_LaserEnvelope.md
MaxThevenet: Thanks, I'll have a look.

EXT_LaserEnvelope.md
MaxThevenet: The goal of this optional parameter is to address the practical case when this file would be shared between different people to run the same simulation with e.g. different codes. There it could be valuable to make sure all the required laser information is contained in the openPMD file, rather than sharing the file + some information like the propagation direction. In practice, this could be used e.g. to set the antenna normal vector. Options are:
- Drop this idea, and also communicated the propagation direction (and potentially other elements that I missed)
- Keep this parameter (and fix the array issue, similar to above)
- Include it in a different way in the array (although I don't see a simple and obvious way)

What do you think?

EXT_LaserEnvelope.md
MaxThevenet: This should be addressed in the last commit.

EXT_LaserEnvelope.md
RemiLehe: Hm, I don't have a very strong opinion, but I guess for now it might be okay to just drop this parameter.
We could also discuss this point further at the next `lasy` meeting.

MaxThevenet: Sounds good, let's remove it, then we can add it again if needed

MaxThevenet:
Thanks for your reviews, I think all comments are addressed now. @ax3l once we agree on this standard we can merge to `upcoming-2.0.0` IMO.

RemiLehe (approved):

EXT_LaserEnvelope.md
```diff
@@ -0,0 +1,47 @@
...
+   E_x &= \operatorname{Re}\left( \mathcal{E} e^{-i\omega_0t}p_x\right)\\
+   E_y &= \operatorname{Re}\left( \mathcal{E} e^{-i\omega_0t}p_y\right)\\
+\end{aligned}
+```
+
+where $\operatorname{Re}$ stands for real part,  $E_x$ (resp. $E_y$) is the laser electric field in the x (resp. y) direction, $\mathcal{E}$ is the complex laser envelope described in the standard, $\omega_0 = 2\pi c/\lambda_0$ is the angular frequency defined from the laser wavelength $\lambda_0$ and $(p_x,p_y)$ is the (complex and normalized) polarization vector. The polarization state (linear, circular, elliptical) is controlled by the phase of the polarization vector. For instance, if $arg(p_x) = arg(p_y)$, the polarization is linear. If $arg(p_x) = arg(p_y) + \pi/2$, the polarization is circular.
+
+When added to an output, the following naming conventions shall be used for complex electric field `mesh records`.
+
+  - `laserEnvelope/`
```
ax3l: Idea discussed today: how can we have multiple laser pulses in one series (e.g. file).

We could keep the name arbitrary but identify the laser envelope by an attribute (e.g., `isLaserEnvelope=true`)

EXT_LaserEnvelope.md
```diff
@@ -0,0 +1,47 @@
...
+    - type: *(complexX)*
+    - scope: *(required)*
+    - decription: Scalar field for the envelope (in V/m). See above for description.
```
ax3l: We can impose V/m even in the data, but we can also keep the usual base-standard definition freedom in the raw data and enforce V/m conversion with `unitSI`.
```suggestion
    - description: Scalar field for the envelope (as a field strength). See above for description.
```

MaxThevenet:
Thanks, I updated the standard to allow for arbitrary name, with `<laser envelope name>` as a placeholder. Would this work for you?

Also, rather than looping over all mesh records and checking whether `isLaserEnvelope == True`, we're considering in HiPACE++ asking the user to specify the name of the laser envelope field (and then check for this field whether `isLaserEnvelope == true`). This is relatively convenient in the code, and avoids cumbersome logic when e.g. a file has multiple laser pulses. Should this be the advised behavior, or do we give no recommendation and let the application developers choose?

ax3l:
Great point to consider.

I think by default, one can pick the first laser pulse found in a provided file.

I think if multiple pulses are in a file, one should either load all of them or warn if a code that can only handle one pulse picks one (e.g., the first found).

If a user picks one by name, then of course that one is taken (no warning). Does that make sense?

MaxThevenet:
Just coming back to this PR. Is there something else needed?

ax3l (approved):
Looks great, in it goes! :)

PR #275, opened 2023-05-19
Replace operatorname with mathrm in LateX MD rendering

MaxThevenet:
This fixes rendering markdown issues in https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/EXT_LaserEnvelope.md.

Follow-up to #271

## Affected Components

- `EXT_LaserEnvelope.md`

PR #276, opened 2023-06-05
Markdown `EXT:LaserEnvelope` LaTeX Fix

ax3l:
## Description

Per Support:
Thanks for reaching out to GitHub Support about problems rendering LaTeX that uses the \operatorname macro. I see that we just removed support for some potentially harmful LaTeX macros that allow for either overriding other macros or operators, or would potentially allow for invisible text. Among these is \operatorname.

Same as https://github.com/LASY-org/lasy/pull/134

Follow-up to #271

## Affected Components

- `EXT-LaserEnvelope`

## Logic Changes

None.

## Writer Changes

None.

## Reader Changes

None.

## Data Converter

None.

### DIFF
--- a/EXT_LaserEnvelope.md
+++ b/EXT_LaserEnvelope.md
@@ -18,12 +18,12 @@ The laser pulse is represented as a single complex field $\mathcal{E}$, describi
 
 ```math
    \begin{aligned}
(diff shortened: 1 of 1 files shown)


ax3l:
Ooops, duplicate of #275

PR #281, opened 2023-08-03
Improvements of the laser envelope extension

MaxThevenet:
This PR proposes the following improvements to the EXT_LaserEnvelope extension:
- `isLaserEnvelope` is a required attribute, rather than optional.
- The standard supports both z and t representations for the longitudinal axis. ~~Attributes `isSpatial` contains this info.~~
- The standard supports both electric field (V/m) and vector potential a0. Attributes `isNormalized` contains this info.

While the first point is minor and the second point has been discussed in the context of LASY, the last point is relatively new. Supporting the normalized vector potential (or at least the SI vector potential) can be important as this is what appears in the ponderomotive force, and what is implemented in envelope solvers. As the conversion from field to vector potential can be non-trivial in a simulation code (it required the knowledge of the local frequency, which comes from a derivative of the phase, and can be cumbersome to implement in a robust way), I believe it is useful for the standard to accept both. An alternative approach would be to add, as an attribute, another array containing the local frequency, so the reconstruction in simulation codes is made easier.


ax3l:
This sounds great. My only thought is, do we want to have a non-binary option instead of `isNormalized`? I think that normalizing by:
- peak V/m
- peak a0
- total energy
- V/m @FWHM
- a0 @ FWHM
- ...

are other options often used in experimental measurements. We could stay extensible if we only standardize a few strings for now: `normalization=`
- `"peak_a0"`
- `"peak_Vm"`
- `"energy_J"`
- `"fwhm_a0"`
- `"fwhm_Vm"`
- ...

RemiLehe:
Hm, I think that the name `isNormalized` might be confusing: it maybe suggests that this just changes the field by a normalizing constant, whereas that is not the case.
The fields that are stored with `isNormalized = false` and `isNormalized = true` are **not** simply proportional to each other - since in one case the field is the electric field and in the other case it is the vector potential.

Maybe we could have something more explicit like, `envelopeField = 'electric_field'` or `envelopeField = 'normalized_vector_potential'`. (Btw, having this single metadata could replace both `isEnvelope` and `isNormalized`.)

MaxThevenet:
@ax3l Thanks, I think that, as Remi pointed out, the name `isNormalized` was confusing: vector potential and electric field are two different things, as seen on the figure below (from a realistic FBPIC simulation, these are temporal profiles, sorry there's no axes). I think we can start with just the 2 options, and maybe add some others later, but maybe it's good to be quite restrictive here, to make the standard lightweight?
Screenshot 2023-08-10 at 09 09 02


@RemiLehe Thanks, this is better indeed. So to check if an openPMD file complies with the `EXT_LaserEnvelope`, one would check if `envelopeField` is present, and which of the 2 possible values it has? I'll make the change, and we can have another round before merging.

MaxThevenet:
I made the changes :)

EXT_LaserEnvelope.md
```diff
@@ -23,7 +23,16 @@ The laser pulse is represented as a single complex field $\mathcal{E}$, describi
...
+```math
+   \begin{aligned}
+   E_x &= \mathrm{Re}\left( \mathcal{E} e^{ik_0z}p_x\right)\\
+   E_y &= \mathrm{Re}\left( \mathcal{E} e^{ik_0z}p_y\right)\\
+\end{aligned}
+```
+
+where $\mathrm{Re}$ stands for real part, $E_x$ (resp. $E_y$) is the laser electric field in the x (resp. y) direction, $\mathcal{E}$ is the complex laser envelope described in the standard, $\omega_0 = 2\pi c/\lambda_0$ is the angular frequency defined from the laser wavelength $\lambda_0$, $k_0$ is the corresponding wavenumber $k_0=2\pi/\lambda_0$ and $(p_x,p_y)$ is the (complex and normalized) polarization vector. The polarization state (linear, circular, elliptical) is controlled by the phase of the polarization vector. For instance, if $arg(p_x) = arg(p_y)$, the polarization is linear. If $arg(p_x) = arg(p_y) + \pi/2$, the polarization is circular. Alternatively, the stored array can represent the normalized vector potential $a_{x/y} = eA_{x/y}/(m_ec)$ where $A_{x/y}$ is the vector potential of the laser, related to the electric field by $E_{x/y} = -\partial_tA_{x/y}$.
 
 When added to an output, the following naming conventions shall be used for complex electric field `mesh records`.
```
RemiLehe, line 38: The field `unitDimension` (a few lines below) should be corrected: this would be `V/m` **only** when `fieldEnvelope == 'electric_field'`.

RemiLehe (approved):

EXT_LaserEnvelope.md
```diff
@@ -7,14 +7,14 @@ openPMD extension name: `LaserEnvelope`
 Introduction
 ------------
 
-This extension is specifically designed for the modeling laser beam propagation.
+This extension is specifically designed for modeling the laser beam propagation.
```
ax3l: Minor idea:
```suggestion
This extension is specifically designed for modeling the propagation of laser pulses.
```

ax3l (approved):

EXT_LaserEnvelope.md
MaxThevenet: Yeah, I think that works well, we should be able to support monochromatic beams.

ax3l (approved):

ax3l:
This is great, thank you! :)

PR #289, opened 2024-08-05
Const Records: Relax `shape` for Particles

ax3l:
## Description

For MPI-parallel I/O output, we developed a new method in ADIOS2 that does not need an initial metadata gather ("JoinedArrays"). To be able to use this mode, we need to relax the requirements to write a shape for constant records in a species (particle group), because otherwise we still have to do a collective gather.

This adds the need for a slight additional read fallback implementation on the reader side.

## Affected Components

- `base`

## Logic Changes

The required attribute `shape` in constant record components is now optional for records in particle groups (species), if there is at least another record to recover the `shape` from in the same particle species.

## Writer Changes

The required attribute `shape` in constant record components is now optional for records in particle groups (species).

## Reader Changes

The required attribute `shape` in constant record components is now optional for records in particle groups (species).
 If the attribute is missing, go through other records and components of the same species and pick the first one that has a `shape` (e.g., non-constant record component full extent or a constant record component with a `shape`) and use that information to recover.

- [ ]  `openPMD-api`: https://github.com/openPMD/openPMD-api/issues/1663 https://github.com/openPMD/openPMD-api/pull/1661 @franzpoeschel 
  - [x]  `yt`: https://github.com/yt-project/yt/... - will be based on openPMD-api soon
  - [x]  `ParaView`: based on openPMD-api
  - [x]  `openPMD-viewer`: base on openPMD-api for ADIOS2
- [ ] `openPMD-validator`: https://github.com/openPMD/openPMD-validator/... @ax3l 
- [ ]  `VisIt`: https://github.com/openPMD/openPMD-visit-plugin/...

## Data Converter

No changes needed. Files from `1.X` will be forward compatible with regards to this change.

### DIFF
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -731,8 +731,10 @@ The record's *data set* `<componentName>` must be replaced with an empty
 
 `shape` is a 1-dimensional array of `N` *(uint64)* elements, where `N` is the
 number of dimensions of the record. It contains the number of elements of each
-dimension that are replaced with a constant value. For `mesh` based records,
-the order of the `N` values must be identical to the axes in `axisLabels`.
+dimension that are replaced with a constant value.
+  - For `mesh` based records, the order of the `N` values must be identical to the axes in `axisLabels`.
+  - For `particle` records, the `shape` attribute can be skipped if there is *at least one more record* in the same particle group (species) that is non-constant or is constant and has the `shape` attribute.
+    Read-logic will then pick any other record to recover the `shape`.
 
 Other required attributes that where previously stored on the *data set* need
 to be added to the new sub-group as well.
--- a/STANDARD.md
+++ b/STANDARD.md
@@ -733,8 +733,8 @@ The record's *data set* `<componentName>` must be replaced with an empty
 number of dimensions of the record. It contains the number of elements of each
 dimension that are replaced with a constant value.
   - For `mesh` based records, the order of the `N` values must be identical to the axes in `axisLabels`.
-  - For `particle` records, the `shape` attribute can be skipped if there is *at least one more record* in the same particle group (species) that is non-constant or is constant and has the `shape` attribute.
-    Read-logic will then pick any other record to recover the `shape`.
+  - The `shape` attribute can be skipped if there is *at least one more record component* in the same mesh record or particle group (species) that is non-constant or is constant and has the `shape` attribute.
+    Read-logic will then pick any other record component of the same mesh (or particle species) to recover the `shape`.
 
 Other required attributes that where previously stored on the *data set* need
 to be added to the new sub-group as well.


franzpoeschel:
This has practical consequences for parsing a corner case of openPMD data, namely constant scalar particle records. 
Take for example this constant scalar particle record:

```
  int32_t   /data/500/particles/e_all/mass/macroWeighted                        attr   = 0                                                          
  uint64_t  /data/500/particles/e_all/mass/shape                                attr   = {1829775}                                                  
  double    /data/500/particles/e_all/mass/timeOffset                           attr   = 0                                                          
  double    /data/500/particles/e_all/mass/unitDimension                        attr   = {0, 1, 0, 0, 0, 0, 0}                                      
  double    /data/500/particles/e_all/mass/unitSI                               attr   = 8.26366e-28                                                
  double    /data/500/particles/e_all/mass/value                                attr   = 0.00110234                                                 
  double    /data/500/particles/e_all/mass/weightingPower                       attr   = 1
```
How does a parser distinguish `/data/500/particles/e_all/mass` from e.g. `/data/500/particles/e_all/momentum`:
```  
  float     /data/500/particles/e_all/momentum/x                          {1829775} = 0 / 0
  float     /data/500/particles/e_all/momentum/y                          {1829775} = 0 / 0
  float     /data/500/particles/e_all/momentum/z                          {1829775} = 0 / 0
```

Currently, the openPMD-api implements this corner case in `ParticleSpecies.cpp` by:
```cpp
            auto value = std::find(att_begin, att_end, "value");
            auto shape = std::find(att_begin, att_end, "shape");
            if (value != att_end && shape != att_end)
            {
                RecordComponent &rc = r;
                IOHandler()->enqueue(IOTask(&rc, pOpen));
                IOHandler()->flush(internal::defaultFlushParams);
                rc.get().m_isConstant = true;
            }
            try
            {
                r.read();
            }
```
I.e., when parsing a group `/data/500/particles/e_all/MYSTERY`, it checks if the group contains the attributes `value` and `shape`. If yes, it's a constant scalar component, otherwise it's a normal record.
With this standard change, I see two options:

1. We only check for the attribute `value`. MIght lead to problems with datasets that for some reason use an attribute named `value`. 
    Follow-up question: For legacy (1.0) datasets, do we keep checking for both attributes? If yes, the relaxed standard definition cannot be applied to old data. If no, some old datasets might be broken.
3. We check if the record contains subgroups or datasets and treat it as a constant component if not. Might lead to problems with partially written data, e.g. after a crash.

I suggest that we decide for one clearly defined scheme to detect constant scalar components and standardize that. 1. is easier to implement.

Also, do I understand it correctly that the relaxed constant component markup is not applicable to Mesh records?

ax3l:
Thanks, good points!

1. I think the `value` descriminator for a constant record component is sufficient, because `value` is a reserved attribute in the openPMD standard and using it without the well-defined meaning is ignoring/violating the standard.
1.1.: yes, for 1.* datasets we will could require `shape`, but technically we could support it already in implementations with low risk of breaks.
2. I would go with 1.

> Also, do I understand it correctly that the relaxed constant component markup is not applicable to Mesh records?

At least for now that is how we define it. I think there is no big benefit in mesh records to support skipping `shape`, but it would be analogous (needs at least another record component present with `shape`).

STANDARD.md
```diff
@@ -731,8 +731,10 @@ The record's *data set* `<componentName>` must be replaced with an empty
 
 `shape` is a 1-dimensional array of `N` *(uint64)* elements, where `N` is the
 number of dimensions of the record. It contains the number of elements of each
-dimension that are replaced with a constant value. For `mesh` based records,
-the order of the `N` values must be identical to the axes in `axisLabels`.
+dimension that are replaced with a constant value.
+  - For `mesh` based records, the order of the `N` values must be identical to the axes in `axisLabels`.
+  - For `particle` records, the `shape` attribute can be skipped if there is *at least one more record* in the same particle group (species) that is non-constant or is constant and has the `shape` attribute.
+    Read-logic will then pick any other record to recover the `shape`.
```
ax3l: Relaxing `shape` for meshes & particles:
```suggestion
  - For `mesh` based records, the order of the `N` values must be identical to the axes in `axisLabels`.
  - The `shape` attribute can be skipped if there is *at least one more record component* in the same mesh record or particle group (species) that is non-constant or is constant and has the `shape` attribute.
    Read-logic will then pick any other record component of the same mesh (or particle species) to recover the `shape`.
```

franzpoeschel (approved):

RemiLehe (approved):

